PR #232, opened 2023-09-04
Spearbit Audit september 2023

0xvv:
## Branch for the spearbit audit

- added `getRiver()` view function to `RedeemManager`
- added `setOperatorStoppedValidatorCount` event 
- added an optional global unlock schedule to TLC schedules (will be merged in this branch soon)

### More info on the global unlock schedule :
Due to legal & regulatory concerns it was to chosen to apply a global unlock schedule to most, but not all vesting schedules.
The global unlock schedule must release 1/24 th of the total amount of the vesting schedule every month for 2 years after the schedule local lock end. The unlocking is entirely separated from the vesting it just caps the withdrawal of the vested tokens.
Some schedules are not subject to the global unlock schedule, this is why we added the `IgnoreGlobalUnlockSchedule` mapping and altered the vesting schedule creation function to fit the new behaviour.
The global unlock schedule computation is done in the existing function where the local lock is taken into account.  

Those changes are accompanied by a migration, contained in the migration contract, which should be applied before pointing to the new implementation.
The migration sets `IgnoreGlobalUnlockSchedule` for the schedules where it's needed and also updates some erroneous values currently set.


cebidhem (approved):
Relevant fixes from Spearbit audit have been implemented.

2 CI tests are failing:
- lint
- check deploy scripts

Both are ok locally, but I don't want to add another commit to fix those now that the Audit period is over. This will need to be fixed after this PR.

In the meantime, LGTM 🚀

PR #238, opened 2023-10-23
Remove TLC code

0xvv:
## Description

The TLC token will be developed in a dedicated repository from now on.

## Notice

## Pull Request Type

- [x] 🕹️ Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)

## Breaking changes (if applicable)

Removed TLC code


iamsahu (commented):
The files related to deployment of TLC are still present in this repo. I think it should be moved to the new TLC repo so that everything related to it are present in one place and not spread across multiple repos.

0xvv:
We discussed with Mehdi leaving yarn behind and having a fully foundry based repo, that means the deployment stack of the TLC repo will be different and i think it makes more sense leaving the existing here since it was deployed using the scripts in this repo and will not necessarily be compatible with future TLC deployments which would cause duplication and confusion imo.
Note that the existing deployment artifacts are duplicated in the TLC repos for ease of access, simply not the deployments scripts and their outputs.

The same way i did not remove TLC related entries from the CHANGELOG of this repository, which i think makes sense

mischat:
@iamsahu do we still want to get this PR merged in ? It is still relevant ? If so, could you please fix the conflict, and lets try and get it merged in

PR #261, opened 2024-03-04
Claim Freeze for denied user (ETH-1618)

iamsahu:
## Description

A user which was denied after they have submitted a redeem request could still claim the request. This fix aims to prevent the denied user from claiming the redeem request. The ETH remains behind in the redeem manage and is not included in the protocol calculation.

## Notice

## Pull Request Type

- [x] 🛠️ Bug fix (Non-breaking Change: Fixes an issue)


mischat (approved):
This LGTM

contracts/src/RedeemManager.1.sol
```diff
@@ -474,6 +476,10 @@ contract RedeemManagerV1 is Initializable, IRedeemManagerV1 {
             // we load the redeem request in memory
             params.redeemRequest = redeemRequests[_redeemRequestIds[idx]];
 
+            if (allowList.isDenied(params.redeemRequest.owner)) {
+                revert ClaimOwnerIsDenied();
+            }
+
```
mischat, line 482: @iamsahu i feel like we might need to also check the params.redeemRequest.receipient against the DENY list ? As otherwise a blacklist wallet can still receive ETH, which doesn't seem like the right thing ?!

mischat (commented):
I have a thought on this, as it relates to https://docs.google.com/document/d/1dpKSXh2u17r0bTAVu8onbN_me-VLNd1pnBjNjSzo9Bk/edit#bookmark=id.hfo981b88a5t

contracts/src/RedeemManager.1.sol
iamsahu: The recipient of the redeem request is the same as the owner of the redeemRequest. So, we don't need a separate check.

PR #273, opened 2024-07-29
Abstract contract for protocol version

iamsahu:
## Description

Since, we are planning to deploy a new version of the protocol (except firewall & proxy) whenever any of the contract has a change to be deployed to the mainnet. Putting versioning in an abstract contract would make it easy to update the version across the contracts with just one change.

## Notice

## Pull Request Type

- [x] 🕹️ Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)

## Testing

- [x] Have you tested this code with the official test suite?

### DIFF
--- a/contracts/src/Allowlist.1.sol
+++ b/contracts/src/Allowlist.1.sol
@@ -2,7 +2,7 @@
 pragma solidity 0.8.20;
 
 import "./interfaces/IAllowlist.1.sol";
-import "./interfaces/IVersion.sol";
+import "./ProtocolVersion.sol";
 
 import "./libraries/LibAllowlistMasks.sol";
 import "./Initializable.sol";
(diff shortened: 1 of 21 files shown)


contracts/src/Allowlist.1.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity 0.8.20;
 
 import "./interfaces/IAllowlist.1.sol";
-import "./interfaces/IVersion.sol";
+import "./ProtocolVersion.sol";
 
 import "./libraries/LibAllowlistMasks.sol";
 import "./Initializable.sol";
```
koderholic, line 8: @iamsahu It seems this import is supposed to be ProtocolVersionV1 everywhere.

koderholic (changes requested):

contracts/src/Allowlist.1.sol
iamsahu: The current import is correct since the file name is the same. We could have used the following format to do the import: `import {ProtocolVersionV1} from "./ProtocolVersion.sol";`
I didn't want to differ from rest of the imports so I have followed the same pattern as we have for the rest.

koderholic: Oh! I see, it's a bit confusing though. Why not change the file name?

koderholic (approved):
LGTM

PR #274, opened 2024-07-31
Update redeemManager to revert claim if initiator is denied [ETH-2118]

koderholic:
## Description

This checks that the original request’s funds owner wasn’t blacklisted, o/w they can create temporary addresses for the recipient address tat have not be blacklisted for the recipient address, and still be able to claim their funds with those addresses.

Changes included:
- Updated the redeemRequests struct to include the initiator of the redeem request
- Updated requestRedeem function to store the msg.sender as the initiator of the redeemRequest
- Updated claimRedeemRequest to revert if the initiator of a redeemRequest is denied

## Notice

- [x] Have you added sufficient documentation in your code?

## Pull Request Type

- [x] 🛠️ Bug fix (Non-breaking Change: Fixes an issue)

### DIFF
--- a/contracts/src/state/redeemManager/RedeemQueue.sol
+++ b/contracts/src/state/redeemManager/RedeemQueue.sol
@@ -15,6 +15,8 @@ library RedeemQueue {
         uint256 maxRedeemableEth;
         /// @custom:attribute The owner of the redeem request
         address owner;
+        /// @custom:attribute The initiator of the redeem request
+        address initiator;
         /// @custom:attribute The height is the cumulative sum of all the sizes of preceding redeem requests
         uint256 height;
     }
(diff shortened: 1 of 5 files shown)


contracts/src/state/redeemManager/RedeemQueue.sol
```diff
@@ -15,6 +15,8 @@ library RedeemQueue {
         uint256 maxRedeemableEth;
         /// @custom:attribute The owner of the redeem request
         address owner;
```
iamsahu, line 17: We should change this to `receiver` as the address represents the receiver of the redeemed ETH. It would make it more readable as well. 
Also, we should change the error `ClaimOwnerIsDenied` to `ClaimReceiverIsDenied`.

iamsahu (changes requested):
Apart from one variable name change the rest looks good. Good job!

contracts/src/state/redeemManager/RedeemQueue.sol
koderholic: I agree, receiver makes more sense. Would update now.

PR #345, opened 2026-02-14
fix: enforce allowlist deny checks on wLsETH transfers

AI-Reviewer-QS:
## Summary

- `WLSETH._transfer()` operated solely on internal `BalanceOf` mappings without consulting the River allowlist, allowing denied users to freely transfer wLsETH via `transfer()` and `transferFrom()` — bypassing the denylist enforcement that River applies to LsETH transfers
- Added deny checks in `_transfer()` that query the allowlist via `River.getAllowlist()` and revert with `Denied()` if either sender or receiver is on the denylist, consistent with River's `_onTransfer()` behavior
- Added `Denied` error to `IWLSETHV1` interface

## Test plan

- [x] `testTransferDeniedSender` — denied sender cannot call `transfer()`
- [x] `testTransferDeniedRecipient` — transfer to denied recipient reverts
- [x] `testTransferFromDeniedSender` — denied sender cannot be transferred from via `transferFrom()`
- [x] `testTransferFromDeniedRecipient` — `transferFrom()` to denied recipient reverts
- [x] `testTransferSucceedsWhenNotDenied` — non-denied users can transfer normally
- [x] `testTransferSucceedsAfterUndeny` — user can transfer after being removed from denylist
- [x] All 31 existing WLSETH tests continue to pass (mock updated with `getAllowlist()`)
- [x] All 6 new deny tests are fuzz tests (1500 runs each)

## Release Notes

* **New Features**
  * Added allowlist deny checks to transfer operations, preventing both senders and recipients from transacting if their addresses are blocked.
  * New `Denied` error signals when an account is denied access to transfers.

* **Tests**
  * Added comprehensive test coverage for allowlist denial and undeny scenarios.

### DIFF
--- a/.github/workflows/Certora.yaml
+++ b/.github/workflows/Certora.yaml
@@ -2,16 +2,12 @@ name: Certora verification
 
 on:
   push:
-    branches:
-      - main
     paths:
       - "contracts/**"
       - "lib/**"
       - ".github/**"
       - "certora/**"
   pull_request:
-    branches:
-      - main
     paths:
       - "contracts/**"
       - "lib/**"
@@ -41,6 +37,12 @@ jobs:
 
       - name: Install certora
         run: pip3 install certora-cli
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '21'
 
       - name: Install solc
         run: |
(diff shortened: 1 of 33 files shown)


mischat:
I like this ... good catch @AI-Reviewer-QS ... 

cc @juliaaschmidt and @iamsahu

mischat:
we have created a new PR from this, thank you. 

This was a good find, it should be noted that wLsETH doesn't exist anywhere other than in our dev environments, good catch nonetheless! : 

https://github.com/liquid-collective/liquid-collective-protocol/pull/357

PR #346, opened 2026-02-16
feat(test) adds more Deposit BYOV Allocation tests

mischat:
What's missing

  1. No test verifies the actual keys returned match the correct operator. Every test checks publicKeys.length and operator.funded, but none asserts that the keys in positions 0-1 belong to operator 0 and keys 2-4 belong to operator 1. The tests prove the right count per operator but not the right keys per operator.

  The testRegularDepositDistribution test does use vm.expectEmit to check FundedValidatorKeys events with the actual key bytes -- that's the one exception
  that validates key content per operator.

  2. No test for uneven/asymmetric multi-operator allocation with content verification. For example: allocate 1 to op0, 8 to op1, 1 to op2, and verify each operator got the correct keys from their own key set.

  3. No test for a large number of operators (e.g., 10+). All tests use 2-5 operators. A test with, say, 15 operators where only 3 get allocations would exercise the linear search in _updateCountOfPickedValidatorsForEachOperator more thoroughly.

  4. No test for allocation to non-contiguous operator indexes. For example, operators exist at indexes 0,1,2,3,4 but allocations only target 0 and 4 (skipping middle ones). The testInactiveDepositDistribution covers this partially (operators 0,2,4) but only when the skipped ones are inactive, not when active operators are simply not allocated to.

  5. No test that passes allocations through the full depositToConsensusLayerWithDepositRoot -> River -> OperatorsRegistry path and verifies per-operator funded counts. The River and DepositManager tests mostly check reverts and total counts, not per-operator correctness through the full stack.

  Bottom line

  The count correctness ("operator X got N validators") is well tested. The content correctness ("the keys returned for operator X are actually operator
  X's keys") relies almost entirely on the FundedValidatorKeys event assertions in testRegularDepositDistribution. I'd recommend at least one more test
  that does a multi-operator asymmetric allocation and verifies the returned key bytes match each operator's registered keys.

## Notice

## Pull Request Type

## Testing

### DIFF
--- a/contracts/test/OperatorsRegistry.1.t.sol
+++ b/contracts/test/OperatorsRegistry.1.t.sol
@@ -4253,7 +4253,8 @@ contract OperatorsRegistryV1AllocationCorrectnessTests is OperatorAllocationTest
             address opAddr = makeAddr(string(abi.encodePacked("op", vm.toString(i))));
             operatorsRegistry.addOperator(string(abi.encodePacked("Operator ", vm.toString(i))), opAddr);
 
-            bytes memory keys = genBytes(uint256(keysPerOp) * (ValidatorKeys.PUBLIC_KEY_LENGTH + ValidatorKeys.SIGNATURE_LENGTH));
+            bytes memory keys =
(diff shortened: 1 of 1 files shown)


juliaaschmidt (approved):
Nice! 👍

PR #372, opened 2026-03-13
Introduction of pectra functions of consolidation & withdraw on the WC (BS-3152)

iamsahu:
## Description

This pull request introduces significant enhancements to the withdrawal contract to support the upcoming Pectra protocol features. The main changes include adding support for validator withdrawals and consolidations, new storage utilities for contract addresses, and improved safety and error handling. These updates enable the contract to interact with external execution layer (EL) contracts for Pectra operations, while ensuring robust validation and fee management.

**Pectra Feature Integration**

* Added `withdraw` and `consolidate` functions to `WithdrawV1`, allowing River to request validator withdrawals and consolidations via external EL contracts, with fee handling and excess refund logic. [[1]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R60-R173) [[2]](diffhunk://#diff-984b02abbb24a29ef36d5b26931fab83ad9df15d29714c4a53269f9f8354b39fR79-R100)
* Introduced new events and error types in `IWithdrawV1` for tracking withdrawal/consolidation requests, fee refunds, and error scenarios.

**Storage Utilities**

* Added `WithdrawalContractAddress` and `ConsolidationContractAddress` libraries for securely storing and retrieving EL contract addresses used in Pectra operations. [[1]](diffhunk://#diff-6bcc4775e204529a288d68b2b5027ef54f74027ca49d110023ba05af5f26c333R1-R26) [[2]](diffhunk://#diff-eb9e9fb427bc4cd4292276eda547b413a9bb9a5997b5bd8c2a661aac4a4df3eeR1-R26)
* Implemented `initWithdrawV1_1` function for initializing EL contract addresses after the contract is deployed. [[1]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R33-R38) [[2]](diffhunk://#diff-984b02abbb24a29ef36d5b26931fab83ad9df15d29714c4a53269f9f8354b39fR8-R67)

**Safety and Validation Improvements**

* Integrated OpenZeppelin's `ReentrancyGuard` to prevent reentrancy attacks on withdrawal and consolidation operations. [[1]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R4-R5) [[2]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R14-R20)
* Added internal validation functions to check pubkey lengths, fee limits, and sufficient value sent, with appropriate error handling and revert logic.

**Versioning**

* Updated contract version to `1.3.0` to reflect these protocol and interface changes.

## Notice

## Pull Request Type

## Testing


contracts/src/Withdraw.1.sol
```diff
@@ -26,6 +30,12 @@ contract WithdrawV1 is IWithdrawV1, Initializable, IProtocolVersion {
         _setRiver(_river);
     }
 
+    /// @inheritdoc IWithdrawV1
+    function initWithdrawV1_1(address _withdrawalContractAddress, address _consolidationContractAddress) external init(1) {
```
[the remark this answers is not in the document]
iamsahu: The contract sits behind a proxy & during the upgrade the function gets called.

contracts/src/interfaces/IOperatorRegistry.1.sol
```diff
@@ -57,6 +67,12 @@ interface IOperatorsRegistryV1 {
     /// @param amount The amount of requested exits in ETH(wei)
     event RequestedETHExits(uint256 indexed index, uint256 amount);
```
juliaaschmidt, line 78: This event and the one below: `RequestedETHExits`  have the same notice description, which does not help. 

Is this event here for full exits? If so, then amount would always be zero for full EL exits?

contracts/src/interfaces/IOperatorRegistry.1.sol
```diff
@@ -137,6 +153,12 @@ interface IOperatorsRegistryV1 {
     /// @param available The available ETH(wei) amount
     error ExitsRequestedExceedAvailableFundedAmount(uint256 operatorIndex, uint256 requested, uint256 available);
 
+    /// @notice The provided exit requests exceed the available funded ETH amount of the operator
```
juliaaschmidt, line 166: Does this error condition make sense? We limit how much one withdrawal can withdraw by the amount that was ever deposited to that validator?

contracts/src/River.1.sol
```diff
@@ -368,6 +368,19 @@ contract RiverV1 is
         }
     }
 
+    /// @inheritdoc IRiverV1
+    function consolidate(IWithdrawV1.ConsolidationRequest[] calldata requests, uint256 maxFeePerConsolidation)
+        external
+        payable
+        onlyKeeper
```
juliaaschmidt, line 296: I wonder whether this should also be exposed to the same Keeper role. Yesterday we mentioned that it could be a completely new/separate role

contracts/src/interfaces/IOperatorRegistry.1.sol
```diff
@@ -164,7 +186,8 @@ interface IOperatorsRegistryV1 {
     // function initOperatorsRegistryV1_1() external;
 
     /// @notice Migrates operators from V2 to V3 storage, dropping key-management fields
-    function initOperatorsRegistryV1_2() external;
+    /// @param _lcWithdrawAddress The address of the LC owned Withdrawal Address
```
juliaaschmidt: Can we drop the LC owned here? and just have it `WithdrawAddress` because all other contract addresses we store in River state for example do not have the LC prefix.
```
import "./state/river/AllowlistAddress.sol";
import "./state/river/AttestationVerifierAddress.sol";
import "./state/river/RedeemManagerAddress.sol";
import "./state/river/OperatorsRegistryAddress.sol";
import "./state/river/CollectorAddress.sol";
import "./state/river/ELFeeRecipientAddress.sol";
import "./state/river/CoverageFundAddress.sol";
```

contracts/src/interfaces/IRiver.1.sol
```diff
@@ -290,4 +303,13 @@ interface IRiverV1 is IConsensusLayerDepositManagerV1, IUserDepositManagerV1, IS
 
     /// @notice Input for the redeem manager funds
     function sendRedeemManagerExceedingFunds() external payable;
+
+    /// @notice Request Pectra consolidations via the Withdraw contract. Callable by admin; fee ETH sent as msg.value.
```
juliaaschmidt: Remove the only callable by admin here. It mentions it is only callable by keeper l. 308.

contracts/src/interfaces/IRiver.1.sol
```diff
@@ -290,4 +303,13 @@ interface IRiverV1 is IConsensusLayerDepositManagerV1, IUserDepositManagerV1, IS
...
+    /// @dev Only callable by the keeper
+    /// @dev Since we consolidate to validators we own there is no need to track the consolidation buffer
+    /// @param requests Consolidation requests (each: src pubkeys -> target pubkey)
```
juliaaschmidt: ConsolidationRequest is actually a nested type of many to one not a 1:1 at the moment

contracts/src/state/withdraw/ConsolidationContractAddress.sol
juliaaschmidt, line 1: I think I would move `/state/shared/ConsolidationContractAddress.sol` and the withdrawal one into its own `state/withdraw/` folder?

contracts/src/state/shared/OperatorsRegistryAddress.sol
juliaaschmidt, line 1: Why was this moved? I only see it being used by River anyways?

contracts/src/interfaces/IWithdraw.1.sol
```diff
@@ -5,13 +5,74 @@ pragma solidity 0.8.34;
...
+    /// @param fee Fee paid to the EL consolidation contract
+    event ConsolidationRequested(bytes srcPubkey, bytes targetPubkey, uint256 fee);
+
+    /// @notice Emitted when excess fee refund could not be sent to the recipient
+    /// @param recipient Intended excess fee recipient
+    /// @param amount Amount that could not be sent
+    event UnsentExcessFee(address recipient, uint256 amount);
+
+    /// @notice Thrown when pubkeys and amounts array lengths differ
+    error LengthMismatch(uint256 pubkeysLength, uint256 amountLength);
```
juliaaschmidt, line 40: input args natspec is missing

contracts/test/River.1.t.sol
```diff
@@ -2878,3 +2879,151 @@ contract RiverV1CoverageTests is RiverV1TestBase {
...
+// --- Mocks for River Pectra (withdraw/consolidate) tests ---
+
+contract MockELWithdrawalForRiver {
+    uint256 public fee = 1 gwei;
+
+    function setFee(uint256 _fee) external {
+        fee = _fee;
+    }
+
+    fallback(bytes calldata) external payable returns (bytes memory) {
```
juliaaschmidt, line 3180: Can we make this mock closer to reality? eg. each consolidation should emit an event encoding the the src and target pub key

contracts/test/River.1.t.sol
```diff
@@ -2878,3 +2879,151 @@ contract RiverV1CoverageTests is RiverV1TestBase {
...
+        );
+        withdraw.initializeWithdrawV1(address(river));
+        mockWithdrawal = new MockELWithdrawalForRiver();
+        mockConsolidation = new MockELConsolidationForRiver();
+        withdraw.initWithdrawV1_1(address(mockWithdrawal), address(mockConsolidation), address(operatorsRegistry));
+        vm.prank(admin);
+        river.setKeeper(admin);
+    }
+
+    function testRiverConsolidateAsAdminEmitsEventAndForwards() public {
```
juliaaschmidt: this is asKeeper, there's no admin. same below and across all the tests. 

If you want a new role, I believe withdrawals should be subject to keeper but consolidations because they could move stake out of the system could be another role.

contracts/test/fork/mainnet/2.operatorsMigrationV1toV3.t.sol
juliaaschmidt, line 1: This is the old migration V1->V3 without the LCwithdraw address. It is not realistic to add it here no?

contracts/test/fork/mainnet/2.operatorsMigrationV2toV3.t.sol
juliaaschmidt, line 1: address(1) is repeated four times in this file, can you please create a variable instead of hardcoding an address everywhere :)                                                                                                    
                                                                                                                                                                                                                                                                         
- it doesn't exercise what production will do. The mainnet upgrade tx will pass the real WithdrawV1 proxy. Hard-coding a synthetic value means this test never catches a bug that would only manifest with an address that actually has code at it 
-the slot-load assertion further down (bytes32(uint256(uint160(address(1))))) is even harder to parse.

contracts/test/OperatorsRegistry.1.t.sol
```diff
@@ -16,9 +16,7 @@ import "../src/state/operatorsRegistry/TotalValidatorExitsRequested.sol";
 
 contract OperatorsRegistryInitializableV1 is OperatorsRegistryV1 {
     /// @dev Override to allow tests to call functions without pranking as river
-    modifier onlyRiver() override {
-        _;
-    }
+    // function _checkRiver() internal override {}
```
juliaaschmidt: I think we can remove these comments

juliaaschmidt (commented):
Great stuff! 👍

contracts/src/River.1.sol
iamsahu: That's right, this is something we need to discuss and come to a conclusion on. cc @mischat

contracts/src/interfaces/IOperatorRegistry.1.sol
iamsahu: `RequestedETHExits`: This is for full CL Exits.
Have updated the comment

contracts/src/interfaces/IOperatorRegistry.1.sol
iamsahu: This is not checking the withdrawal per validator but per NO & what is available on CL for exit. Since, it is a NO level check I think it makes sense.

contracts/src/interfaces/IRiver.1.sol
iamsahu: Removed

contracts/src/interfaces/IRiver.1.sol
iamsahu: Fixed

contracts/src/state/withdraw/ConsolidationContractAddress.sol
iamsahu: Makes sense, will do.

contracts/src/state/shared/OperatorsRegistryAddress.sol
iamsahu: Its being used by withdraw contract as well.

contracts/src/interfaces/IWithdraw.1.sol
iamsahu: Added.

contracts/test/River.1.t.sol
iamsahu: Fixed

iamsahu: Will introduce a new role once we come to a conclusion on it.

contracts/src/interfaces/IOperatorRegistry.1.sol
iamsahu: Done

contracts/src/Withdraw.1.sol
```diff
@@ -47,13 +63,131 @@ contract WithdrawV1 is IWithdrawV1, Initializable, IProtocolVersion {
...
+    ) external payable onlyRiver nonReentrant {
+        if (requests.length == 0) {
+            revert InvalidEmptyArray();
+        }
+        address consolidationContract = PectraConsolidationContractAddress.get();
+        uint256 fee = _validateAndReturnFee(consolidationContract, maxFeePerConsolidation);
+
+        uint256 totalNumOfConsolidationOperations = 0;
+        for (uint256 i = 0; i < requests.length; i++) {
+            if (requests[i].srcPubkeys.length == 0) {
```
juliaaschmidt, line 118: we should also fail early on no target pubkeys  `(.....|| requests[i].targetPubkeys.length is == 0)` no?

contracts/src/interfaces/IOperatorRegistry.1.sol
```diff
@@ -281,7 +309,13 @@ interface IOperatorsRegistryV1 {
     /// @dev Reverts with ExitsRequestedExceedExitDemand if total exits requested exceed the current demand
     /// @dev Reverts with NoExitRequestsToPerform if there is no pending exit demand
     /// @param _allocations The proposed per-operator exit ETH allocations, sorted by operator index
-    function requestETHExits(ExitETHAllocation[] calldata _allocations) external;
+    /// @param _partialAllocations The proposed per-operator per-pubkey partial exit ETH allocations, sorted by operator index
+    /// @param _maxFeePerWithdrawal The maximum fee for per withdrawal request
+    function requestETHExits(
```
juliaaschmidt, line 314: maybe would be clearer if it was `requestELExits` ?

contracts/src/OperatorsRegistry.1.sol
```diff
@@ -253,7 +256,11 @@ contract OperatorsRegistryV1 is IOperatorsRegistryV1, Initializable, Administrab
...
 
     /// @inheritdoc IOperatorsRegistryV1
-    function requestETHExits(ExitETHAllocation[] calldata _allocations) external {
+    function requestETHExits(
+        ExitETHAllocation[] calldata _allocations,
+        PartialExitETHAllocation[] calldata _partialAllocations,
+        uint256 _maxFeePerWithdrawal
+    ) external payable {
         if (msg.sender != IConsensusLayerDepositManagerV1(RiverAddress.get()).getKeeper()) {
             revert IConsensusLayerDepositManagerV1.OnlyKeeper();
```
juliaaschmidt, line 265: Should this actually be fixed to keeper or set to a separate role like `consolidate()` but could be set to the same address?

contracts/src/River.1.sol
juliaaschmidt: As per discussion, separate new role to give us flexibility. The new role might be set to keeper in the backend.

contracts/src/Withdraw.1.sol
iamsahu: No, since `targetPubKey` is not an array.

contracts/src/OperatorsRegistry.1.sol
iamsahu: We can introduce in another PR.

contracts/src/River.1.sol
mischat: happy to create a ticket to close this out, and have the change come in a smaller PR

mischat: https://github.com/liquid-collective/liquid-collective-protocol/issues/523

contracts/src/Withdraw.1.sol
```diff
@@ -9,11 +11,14 @@ import "./libraries/LibErrors.sol";
 import "./libraries/LibUint256.sol";
 
 import "./state/shared/RiverAddress.sol";
+import "./state/shared/OperatorsRegistryAddress.sol";
+import "./state/withdraw/PectraWithdrawalContractAddress.sol";
+import "./state/withdraw/PectraConsolidationContractAddress.sol";
 
 /// @title Withdraw (v1)
 /// @author Alluvial Finance Inc.
 /// @notice This contract is in charge of holding the exit and skimming funds and allow river to pull these funds
```
mischat, line 20: we should probably update this in a future ticket too.

contracts/src/Withdraw.1.sol
```diff
@@ -47,13 +63,131 @@ contract WithdrawV1 is IWithdrawV1, Initializable, IProtocolVersion {
...
         emit SetRiver(_river);
     }
 
+    /// @notice Internal: validate fee from EL contract and ensure it does not exceed max
+    function _validateAndReturnFee(address feeContract, uint256 _maxAllowedFee) internal view returns (uint256 _fee) {
+        (bool readOK, bytes memory feeData) = feeContract.staticcall("");
+        if (!readOK) {
+            revert FeeReadFailed();
+        }
+        _fee = abi.decode(feeData, (uint256));
```
mischat, line 157: 2. Fee decoding differs
  2. Fee decoding differs

  - LC (:157): _fee = abi.decode(feeData, (uint256));
  - TVS (:219): _fee = uint256(bytes32(feeData));

  Both expect the EIP-7002/7251 predeploy to return a 32-byte fee. abi.decode is stricter: it reverts if feeData is shorter than 32 bytes. The TVS bytes32(feeData) cast right-pads/truncates instead. For the real predeploys returning exactly 32 bytes they're equivalent, but LC's form is the safer of the two. 

should we change this in TVS ? which one is cleaner ?

contracts/src/Withdraw.1.sol
```diff
@@ -47,13 +63,131 @@ contract WithdrawV1 is IWithdrawV1, Initializable, IProtocolVersion {
...
+    function _validatePubkeyLength(bytes calldata pubkey) internal pure {
+        if (pubkey.length != 48) {
+            revert InvalidPubkeyLength(pubkey.length);
+        }
+    }
+
+    /// @notice Internal: refund excess fee to recipient; emit on send failure instead of reverting
+    function _refundExcessFee(uint256 _totalValueReceived, uint256 _totalFeePaid, address _excessFeeRecipient)
+        internal
+    {
```
mischat, line 181: Missing zero-address guard on _excessFeeRecipient — silent loss of funds.

  If _excessFeeRecipient == address(0) and there's excess value, the refund call{value: excess}("") to the zero address succeeds, so the excess is burned rather than
  returned — and because send failure only emits UnsentExcessFee instead of reverting, there's no signal that anything went wrong.

  The TVS implementation guards against this at the top of the same function:
```
  if (_excessFeeRecipient == address(0)) revert InvalidAddress();
```

  Suggest adding the equivalent check here:
```
  function _refundExcessFee(uint256 _totalValueReceived, uint256 _totalFeePaid, address _excessFeeRecipient)
      internal
  {
      if (_excessFeeRecipient == address(0)) {
          revert InvalidAddress();
      }
```

contracts/src/state/operatorsRegistry/WithdrawAddress.sol
```diff
@@ -0,0 +1,25 @@
...
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+import "../../libraries/LibSanitize.sol";
+
+/// @title Withdraw Address Storage
+/// @notice Utility to manage the Withdraw Address in storage
+library WithdrawAddress {
+    /// @notice Storage slot of the Withdraw Address
+    bytes32 internal constant WITHDRAW_ADDRESS_SLOT = bytes32(uint256(keccak256("river.state.withdrawAddress")) - 1);
```
mischat, line 11: how is this different to contracts/src/state/withdraw/PectraWithdrawalContractAddress.sol ?

Is this the withdrawalCredential?

Either way i think we should update the @title here ...

contracts/src/OperatorsRegistry.1.sol
mischat: i created an issue for this ...

contracts/src/OperatorsRegistry.1.sol
```diff
@@ -253,7 +256,11 @@ contract OperatorsRegistryV1 is IOperatorsRegistryV1, Initializable, Administrab
     }
 
     /// @inheritdoc IOperatorsRegistryV1
-    function requestETHExits(ExitETHAllocation[] calldata _allocations) external {
+    function requestETHExits(
```
mischat, line 259: this should be nonReentrant

 Both hand control to the keeper before _setCurrentETHExitsDemand/_setTotalETHExitsRequested run (:295-296). Worse, it's a read-modify-write: currentETHExitsDemand is captured at :268 and written back at :296. A reentrant call during either refund sees the stale (un-decremented) demand and can over-request; when the outer call resumes it overwrites the inner's decrement with capturedDemand - outerRequested, losing the inner consumption and corrupting CurrentETHExitsDemand.

Perhaps this is NOT a thing, cc @iamsahu ?

contracts/test/River.1.t.sol
```diff
@@ -22,6 +22,7 @@ import "../src/interfaces/IRiver.1.sol";
 import "../src/interfaces/IDepositContract.sol";
 import "../src/interfaces/IDepositDataBuffer.sol";
 import "../src/Withdraw.1.sol";
+import "../src/interfaces/IWithdraw.1.sol";
```
mischat, line 25:
```
  Significant gaps:

  ┌─────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                     Gap                     │                                                      Detail                                                      │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ All fee/value logic                         │ Every partial test calls requestETHExits(empty, allocs, 0) — maxFee = 0 and no msg.value. So fee forwarding, the │
  │                                             │  excess-refund branch (:286-292), and UnsentRefund have zero coverage. This is the part that moves real ETH.     │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │                                             │ MockWithdrawForPartialExits skips all of WithdrawV1's validations. testPartialExitEmptyAmountsIsNoOp passes      │
  │ Mock diverges from real Withdraw            │ empty pubkeys/amounts and asserts a no-op — but the real Withdraw.withdraw reverts InvalidEmptyArray on empty    │
  │                                             │ pubkeys. That test encodes production-invalid behavior. No test wires the real WithdrawV1 into the registry.     │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │                                             │ Not referenced by any test (tests use sudoSetWithdrawAddress). The actual migration path that sets               │
  │ initOperatorsRegistryV1_2(_withdrawAddress) │ WithdrawAddress is uncovered — verify the V2→V3 migration fork tests were updated to the new signature (and      │
  │                                             │ still compile).                                                                                                  │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Mixed full + partial in one call            │ No test passes both arrays non-empty — the combined requestedETHAmount vs demand path and same-operator-in-both  │
  │                                             │ reservation are untested.                                                                                        │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Partial AllocationWithIncorrectAmount       │ The per-amount gweiAmount < 1 gwei revert (:352) is untested (only the full-exit < MIN_ETH_AMOUNT is).           │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Event assertion                             │ RequestedPartialETHExits is never asserted via expectEmit.                                                       │
  ├─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Length mismatch                             │ pubkeys.length != amounts.length (reverts in real Withdraw) untested.                                            │
```

mischat: perhaps this can go to another PR too ? this can be turned into an issue.

mischat (approved):
I have approved, but i believe that we should go through the open comments and make sure that we either create new issues or that we mark them as N/A

contracts/src/OperatorsRegistry.1.sol
```diff
@@ -263,53 +270,125 @@ contract OperatorsRegistryV1 is IOperatorsRegistryV1, Initializable, Administrab
...
+    /// @dev Performs the available-ETH check and updates the operator's `requestedExits`.
+    /// @dev Reverts with the partial- or full-exit specific error depending on `_isPartial`.
+    /// @param _operator Storage reference to the operator whose exit is being reserved
+    /// @param _operatorIndex The operator index (used in revert data)
+    /// @param _amount The ETH(wei) amount to reserve
+    /// @param _isPartial True for partial exits, false for full exits
+    function _reserveOperatorExit(
+        OperatorsV3.Operator storage _operator,
+        uint256 _operatorIndex,
+        uint256 _amount,
```
juliaaschmidt, line 379: surely the amount would already indicate whether it is partial or full EL exit? do we need the boolean here? that risks having the arguments out of sync (amount = 0, boolean = false)

contracts/src/interfaces/IOperatorRegistry.1.sol
juliaaschmidt: I don't see you lowering the funded amount if you have already withdrawn from that node operator though? maybe I'm missing something

juliaaschmidt: ok, can we rename it then, as per

```
 Why iamsahu's reply is half-right

  He says "checking … what is available on CL for exit" — that's accurate to the code. But
  the error name and the natspec both say "Available Funded Amount", which is what made you
   (correctly) doubt it. Both the responder and you are reasoning from a name that doesn't
  match the implementation.

  The real issue — naming/docs, not logic

  This is a meaningful nit worth raising as a thread in the PR. The misleading surfaces
  are:

  1. IOperatorRegistry.1.sol:160 — /// @notice The provided exit requests exceed the
  available funded ETH amount of the operator
  2. IOperatorRegistry.1.sol:164 — error ExitsRequestedExceedAvailableFundedAmount(...)
  3. IOperatorRegistry.1.sol:166-170 — same notice +
  PartialExitsRequestedExceedAvailableFundedAmount
  4. IOperatorRegistry.1.sol:302 — /// @dev Each allocation's ethAmount must be non-zero
  and not exceed the operator's available funded-but-not-yet-exited ETH amount
  5. IOperatorRegistry.1.sol:308 — /// @dev Reverts with
  ExitsRequestedExceedAvailableFundedAmount if count exceeds funded minus requestedExits
  for an operator — this dev comment is flatly wrong; the check is activeCLETH -
  (requestedExits - exitedETH), not funded - requestedExits.

  Suggested rename to match the actual semantics:
  ExitsRequestedExceedAvailableActiveCLAmount (and partial variant), plus rewriting the
  natspec to "active CL ETH minus pending exits".

```

contracts/src/OperatorsRegistry.1.sol
iamsahu: We need to do this as we have to differentiate between CL & EL exits. We cannot rely on the value of amount to determine if its EL or CL exit in all cases. In my opinion there is no risk with the arguments as the boolean is hardcoded in the functions it is called from.

contracts/test/fork/mainnet/2.operatorsMigrationV1toV3.t.sol
iamsahu: We will mark it as deprecated.

contracts/test/fork/mainnet/2.operatorsMigrationV2toV3.t.sol
iamsahu: This was fixed

contracts/src/Withdraw.1.sol
iamsahu: Have created an issue

contracts/src/Withdraw.1.sol
iamsahu: Have created an issue.

contracts/src/OperatorsRegistry.1.sol
iamsahu: Have created an issue for this

contracts/test/River.1.t.sol
iamsahu: Have created an issue for this.

juliaaschmidt (approved):

contracts/src/OperatorsRegistry.1.sol
mischat: @juliaaschmidt

contracts/test/fork/mainnet/2.operatorsMigrationV1toV3.t.sol
```diff
@@ -75,13 +75,23 @@ contract OperatorsMigrationV1ToV2 is Test {
 
         OperatorsRegistryV1 or = OperatorsRegistryV1(OPERATORS_REGISTRY_MAINNET_ADDRESS);
 
-        // Run V1_2 migration (V2 → V3 struct, drops limit/keys/latestKeysEditBlockNumber).
-        // The implementation is already upgraded above; initOperatorsRegistryV1_2 is `onlyAdmin`,
-        // so it must be called from the registry admin (distinct from the proxy admin).
-        vm.prank(OperatorsRegistryV1(address(orProxy)).getAdmin());
-        OperatorsRegistryV1(address(orProxy)).initOperatorsRegistryV1_2();
+        // Run V1_2 migration (V2 → V3 struct, drops limit/keys/latestKeysEditBlockNumber)
```
juliaaschmidt, line 78: @mischat

PR #396, opened 2026-03-30
feat: deposit flow with front-run mitigation and BLS signature verification (BS-3417)

juliaaschmidt:
## Deposit flow with Frontrun Mitigation & BLS signature verification

This pull request introduces major enhancements to the consensus layer deposit flow, adding support for attestation-based deposits and associated administration. The changes enable secure, multi-attester validation of deposit data via BLS signatures, and provide new mechanisms for managing attesters and deposit thresholds. The update also refactors and extends the River and ConsensusLayerDepositManager contracts to support these features, including new view and admin functions, and a robust deposit execution pipeline.

**Attestation-Based Deposit Support:**

- Added a new deposit function, `depositToConsensusLayerWithAttestation`, which validates a quorum of attester signatures (BLS) over deposit data, checks withdrawal credentials, and executes deposits only if the attestation threshold is met. This function also updates operator-funded validator accounting and emits relevant events.
- Implemented parsing and validation of operator indices from deposit metadata, ensuring correct attribution of deposits to operators.

**Attester and Threshold Administration:**

- Introduced admin-only functions to set the deposit data buffer address, manage attester membership, and update the attestation threshold, with appropriate validation and event emission.
- Added view functions to retrieve the deposit data buffer address, attestation threshold, attester count, and attester status.

**River Contract Initialization and Buffer Handling:**

- Added the `initRiverV1_4` initializer to configure the deposit data buffer, attesters, and attestation threshold, including validation of inputs and event emission.
- Implemented `_updateFundedValidatorsFromBuffer` to aggregate deposit amounts by operator and emit per-operator events, supporting the new buffer-based deposit flow.

**Internal Refactoring and Integration:**

- Refactored `ConsensusLayerDepositManagerV1` to inherit from `DepositToConsensusLayerValidation` and implement required storage hooks, integrating attestation and buffer logic throughout the contract. [[1]](diffhunk://#diff-7b338638aeefb6f1e264218c49aab994b7c0b9b06328f2d95b74a82b9a9cc702R20-R40) [[2]](diffhunk://#diff-7b338638aeefb6f1e264218c49aab994b7c0b9b06328f2d95b74a82b9a9cc702R53-R97)
- Added new imports for libraries and state modules required for attestation, BLS signature verification, and buffer management. [[1]](diffhunk://#diff-34420a68f2d8aba02b8bc93ba4306dcb459547b0c3a4815dfa1737a16c14665eR20-R28) [[2]](diffhunk://#diff-7b338638aeefb6f1e264218c49aab994b7c0b9b06328f2d95b74a82b9a9cc702L5-R11) [[3]](diffhunk://#diff-7b338638aeefb6f1e264218c49aab994b7c0b9b06328f2d95b74a82b9a9cc702R20-R40)

These changes collectively enable a more secure and flexible deposit pipeline, supporting multi-attester validation and robust operator accounting.

## Notice

## Pull Request Type

## Testing

## Release Notes

* **New Features**
  * Added attestation-based validator deposit functionality with multi-signature validation support
  * Introduced deposit data buffer management for organizing validator deposit batches
  * Enabled configurable attester quorum system with threshold enforcement

* **Bug Fixes**
  * Enhanced deposit validation to prevent withdrawal credential mismatches

* **Tests**
  * Added comprehensive test coverage for attestation-based deposit flows and validation scenarios


contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,260 @@
...
+            dc.deposit{value: deposits[i].amount}(
+                deposits[i].pubkey,
+                deposits[i].withdrawalCredentials,
+                deposits[i].signature,
+                deposits[i].depositDataRoot
+            );
+        }
+
+        emit DepositsExecuted(depositDataBufferId, depositRootHash, deposits.length);
+    }
```
iamsahu: To preserve the contract size we should not include this in the contract. This is also an unsafe function as it is not permissioned.

contracts/src/River.1.sol
```diff
@@ -337,6 +378,55 @@ contract RiverV1 is
         }
     }
 
+    /// @notice Overridden handler to update operator funded ETH accounting for attestation-based deposits.
+    ///         Aggregates deposit objects by operator index and calls _incrementFundedETH.
+    /// @param deposits Array of deposit objects from the DepositDataBuffer
+    function _updateFundedValidatorsFromBuffer(IDepositDataBuffer.DepositObject[] memory deposits) internal override {
```
iamsahu: We should change this name to reflect that it updates the ETH amount.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -15,18 +17,27 @@ import "../state/river/CommittedBalance.sol";
...
     /// @notice Size of a BLS Public key in bytes
     uint256 public constant PUBLIC_KEY_LENGTH = 48;
     /// @notice Size of a BLS Signature in bytes
     uint256 public constant SIGNATURE_LENGTH = 96;
     /// @notice Size of a deposit in ETH
     uint256 public constant DEPOSIT_SIZE = 32 ether;
 
+    /// @dev ASCII bytes for "operator:" prefix used in metadata encoding
+    bytes9 internal constant OPERATOR_PREFIX = "operator:";
+
```
iamsahu, line 40: Considering we would always be passing in operator index, do we want to add the prefix?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -39,6 +50,51 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+    }
+
+    function _setThreshold(uint256 value) internal override {
+        AttestationThreshold.set(value);
+    }
+
+    /// @dev Compute domain separator dynamically so address(this) resolves to the proxy address.
+    function _domainSeparator() internal view override returns (bytes32) {
+        return keccak256(abi.encode(EIP712_DOMAIN_TYPEHASH, NAME_HASH, VERSION_HASH, block.chainid, address(this)));
+    }
```
iamsahu: Since, the address of the proxy is fixed in LC in context of this particular contract maybe we should opt for an immutable variable for this and not calculate it every time.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -82,78 +142,167 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        return Attesters.isAttester(attester);
+    }
+
+    // -----------------------------------------------------------------------
+    // Admin setters
+    // -----------------------------------------------------------------------
+
+    /// @notice Set the DepositDataBuffer contract address. Admin only.
+    function setDepositDataBuffer(address _depositDataBuffer) external {
+        if (msg.sender != _getRiverAdmin()) revert LibErrors.Unauthorized(msg.sender);
```
iamsahu: Since we are using this check in multiple places we should convert this into a modifier.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -82,78 +142,167 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        bytes[] calldata signatures,
+        BLS12_381.DepositY[] calldata depositYs
+    ) external {
+        // 1. Keeper check
+        if (msg.sender != KeeperAddress.get()) {
+            revert OnlyKeeper();
         }
 
+        // 2. Check withdrawal credentials (cheap SLOAD — fail fast before expensive BLS work)
         bytes32 withdrawalCredentials = WithdrawalCredentials.get();
```
iamsahu, line 238: nit: should it be singular?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -82,78 +142,167 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        bool hasDigit = false;
+        for (uint256 i = 9; i < 32; i++) {
+            uint8 c = uint8(bytes1(metadata << (i * 8)));
+            if (c == 0) break; // null terminator
+            if (c < 0x30 || c > 0x39) revert InvalidOperatorMetadata(metadata); // not ASCII digit
+            operatorIndex = operatorIndex * 10 + (c - 0x30);
+            hasDigit = true;
+        }
+        if (!hasDigit) revert InvalidOperatorMetadata(metadata);
     }
```
iamsahu, line 312: We can potentially eliminate this if we agree that the value passed in the metadata would always be an operator index.

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,260 @@
...
+        // 2. EIP-712 digest
+        bytes32 structHash = keccak256(abi.encode(ATTEST_TYPEHASH, depositDataBufferId, depositRootHash));
+        bytes32 digest = ECDSA.toTypedDataHash(_domainSeparator(), structHash);
+
+        // 3. Verify signatures, check attester set, reject duplicates
+        uint256 validCount = 0;
+        uint256 sigLen = signatures.length;
+        address[] memory seen = new address[](sigLen);
+
+        for (uint256 i = 0; i < sigLen; i++) {
```
iamsahu, line 263: We can potentially fail early here if we introduce a `sigLen<thresh` check here. Before the for loop begins

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,260 @@
...
+        // 6. Check depositYs count
+        if (depositYs.length != deposits.length) {
+            revert BLSSignatureCountMismatch(deposits.length, depositYs.length);
+        }
+
+        // 7. Verify deposit contract root
+        bytes32 onChainRoot = _depositContract().get_deposit_root();
+        if (onChainRoot != depositRootHash) {
+            revert DepositRootMismatch(depositRootHash, onChainRoot);
+        }
```
iamsahu: Can we move this after step 1 so that we could fail early?

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,260 @@
...
+
+        // 8. Verify BLS signatures
+        for (uint256 i = 0; i < deposits.length; i++) {
+            bytes32 wc = abi.decode(deposits[i].withdrawalCredentials, (bytes32));
```
iamsahu: Should we move this outside the loop & pass in the WC in the function call? I understand that we are reading it from the deposit data each time as it would useful in case of non-lc deposits, but maybe passing it in one time would be more effective in case of LC? Let me know your thoughts.

contracts/src/libraries/BLS12_381.sol
```diff
@@ -0,0 +1,326 @@
...
+            // x1 from sig96[0..47], zero-padded to 64 bytes at bytes 80..127
+            mcopy(add(g2pt, 0x70), sig96.offset, 48)
+            // y0 zero-padded to 64 bytes at bytes 144..191
+            mcopy(add(g2pt, 0xb0), add(y0_48, 0x20), 48)
+            // y1 zero-padded to 64 bytes at bytes 208..255
+            mcopy(add(g2pt, 0xf0), add(y1_48, 0x20), 48)
+        }
+        // Clear compression flag bits in first byte of x1 (byte 80 of g2pt)
+        g2pt[80] = bytes1(uint8(g2pt[80]) & 0x1f);
+    }
```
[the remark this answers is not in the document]
iamsahu: Can you come up with an example tests that demonstrates the issue?

contracts/src/state/river/AttestationThreshold.sol
```diff
@@ -0,0 +1,17 @@
+//SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+
+library AttestationThreshold {
```
iamsahu: Natspec please

contracts/src/state/river/Attesters.sol
```diff
@@ -0,0 +1,33 @@
...
+    function setAttester(address account, bool value) internal {
+        bytes32 slot = keccak256(abi.encode(ATTESTER_MAPPING_BASE_SLOT, account));
+        LibUnstructuredStorage.setStorageBool(slot, value);
+    }
+
+    function getCount() internal view returns (uint256) {
+        return LibUnstructuredStorage.getStorageUint256(ATTESTER_COUNT_SLOT);
+    }
+
+    function setCount(uint256 count) internal {
```
iamsahu, line 40: Natspec on the functions please 🙏🏾

contracts/src/state/river/DepositDataBufferAddress.sol
```diff
@@ -0,0 +1,17 @@
+//SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+
+library DepositDataBufferAddress {
```
iamsahu, line 8: Natspec please

contracts/src/state/river/DepositDomainValue.sol
```diff
@@ -0,0 +1,16 @@
+//SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+
+library DepositDomainValue {
```
iamsahu, line 10: Natspec please

contracts/src/River.1.sol
```diff
@@ -136,6 +143,40 @@ contract RiverV1 is
...
+        }
+
+        if (_threshold == 0) revert ZeroThreshold();
+        uint256 attesterCount = Attesters.getCount();
+        if (_threshold > attesterCount) {
+            revert ThresholdExceedsAttesterCount(_threshold, attesterCount);
+        }
+        if (_threshold > MAX_SIGNATURES) {
+            revert ThresholdExceedsMaxSignatures(_threshold, MAX_SIGNATURES);
+        }
```
iamsahu: Can we move this after the `_depositDataBuffer` address check to fail early? Also the check `if (_threshold == 0) revert ZeroThreshold();` on line 168

iamsahu (commented):
Almost to the finish line. Some comments on optimization & failing fast.

contracts/src/libraries/BLS12_381.sol
```diff
@@ -0,0 +1,326 @@
...
+    // -----------------------------------------------------------------------
+    // Public functions
+    // -----------------------------------------------------------------------
+
+    /// @notice Compute the BLS deposit domain for a given genesis fork version.
+    /// @dev domain = DOMAIN_DEPOSIT || sha256(genesisForkVersion || bytes28(0))[0:28]
+    function computeDepositDomain(bytes4 genesisForkVersion) internal pure returns (bytes32) {
+        bytes32 forkDataRoot = sha256(abi.encodePacked(genesisForkVersion, bytes28(0)));
+        return bytes32(DEPOSIT_DOMAIN_TYPE) | (forkDataRoot >> 32);
+    }
```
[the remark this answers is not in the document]
iamsahu: Can you come up with an example tests that demonstrates the issue?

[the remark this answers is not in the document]
iamsahu: Can you tell me from where did you read the ethereum spec from?

contracts/src/libraries/BLS12_381.sol
```diff
@@ -0,0 +1,326 @@
...
+        hex"00000000000000000000000000000000114d1d6855d545a8aa7d76c8cf2e21f267816aef1db507c96655b9d5caac42364e6f38ba0ecb751bad54dcd6b939c2ca";
+
+    /// @dev Lower 32 bytes of G2 cofactor h_eff (used as G2MSM scalar for cofactor clearing).
+    ///      Full h_eff ≈ 0x0bc69f08f2ee75b3584c6a0ea91b352888e2a8e9145ad7689986ff031508ffe1329c008d4b7d31a6dcfe6a67def9d0cb72
+    ///      Lower 32 bytes (bytes 16-47): 0x29c008d4b7d31a6dcfe6a67def9d0cb72... placeholder used for compilation
+    bytes32 internal constant G2_H_EFF_LO = 0x29c008d4b7d31a6dcfe6a67def9d0cb700000000000000000000000000000000;
+
+    /// @dev Upper 16 bytes of G2 cofactor h_eff (right-padded to 32 bytes for G2MSM scalar)
+    bytes32 internal constant G2_H_EFF_HI = 0x0000000000000000000000000000000008ffe1329c008d4b0000000000000000;
+
```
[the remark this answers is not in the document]
iamsahu: Can you cite the spec you used. And also can you create tests that proves the bug

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: We discussed this in our call with @mischat and decided that it should be an optional parameter of the attestation. Thus, if none is provided (empty, 0), then the option is not used, and if it is operator:0 then we mean operator index zero remember.

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: The protocol uses withdrawalCredentials everywhere incl for the singular so I just kept it :)

contracts/src/libraries/BLS12_381.sol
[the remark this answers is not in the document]
juliaaschmidt: Extracted this into a separate issue (https://github.com/liquid-collective/liquid-collective-protocol/issues/408) and Prafful is working on investigating these bugs in his PR: #405

contracts/src/libraries/BLS12_381.sol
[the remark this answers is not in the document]
juliaaschmidt: @coderabbitai addressed in commit  ef131439f02893c70bb0dea5ac4ce2b6fa830e5d

contracts/src/components/DepositToConsensusLayerValidation.sol
juliaaschmidt: We cannot assume that all withdrawalCredentials will be the same because the depositDataBuffer contract is supposed to be universal and usable by other products too.

contracts/src/interfaces/components/IConsensusLayerDepositManager.1.sol
```diff
@@ -75,11 +94,27 @@ interface IConsensusLayerDepositManagerV1 {
...
+    /// @param depositDataBufferId  Batch identifier in the DepositDataBuffer
+    /// @param depositRootHash      Current deposit contract root hash co-signed by attesters
+    /// @param signatures           EIP-712 signatures from attesters
+    /// @param depositYs            Y-coordinates for BLS decompression, one per deposit
+    function depositToConsensusLayerWithAttestation(
+        bytes32 depositDataBufferId,
+        bytes32 depositRootHash,
+        bytes[] calldata signatures,
+        BLS12_381.DepositY[] calldata depositYs
     ) external;
```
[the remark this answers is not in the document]
juliaaschmidt: as discussed with @iamsahu can ignore this script change. The protocol is already deployed

contracts/src/libraries/BLS12_381.sol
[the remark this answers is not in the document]
juliaaschmidt: created an issue for this: https://github.com/liquid-collective/liquid-collective-protocol/issues/409

juliaaschmidt:
@coderabbitai review

contracts/src/libraries/BLS12_381.sol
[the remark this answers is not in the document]
juliaaschmidt: Has this issue been resolved now @coderabbitai

contracts/src/state/river/DomainSeparator.sol
```diff
@@ -0,0 +1,25 @@
...
+    /// @return The cached domain separator
+    function get() internal view returns (bytes32) {
+        return LibUnstructuredStorage.getStorageBytes32(DOMAIN_SEPARATOR_SLOT);
+    }
+
+    /// @notice Cache the domain separator
+    /// @param newValue The new domain separator
+    function set(bytes32 newValue) internal {
+        LibUnstructuredStorage.setStorageBytes32(DOMAIN_SEPARATOR_SLOT, newValue);
+    }
```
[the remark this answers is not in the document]
juliaaschmidt: added a comment here: https://github.com/liquid-collective/liquid-collective-protocol/pull/396/commits/805a14c04986facb06b1a6345e7f8425ef5b9fe5 this is not a major issue.

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,265 @@
...
+            revert InsufficientAttestations(sigLen, _threshold());
+        }
+
+        bytes32 onChainRoot = _depositContract().get_deposit_root();
+        if (onChainRoot != depositRootHash) {
+            revert DepositRootMismatch(depositRootHash, onChainRoot);
+        }
+
+        bytes32 structHash = keccak256(abi.encode(ATTEST_TYPEHASH, depositDataBufferId, depositRootHash));
+        bytes32 digest = ECDSA.toTypedDataHash(_domainSeparator(), structHash);
```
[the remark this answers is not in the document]
juliaaschmidt: addressed in https://github.com/liquid-collective/liquid-collective-protocol/pull/396/commits/805a14c04986facb06b1a6345e7f8425ef5b9fe5

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: Maybe it should be a helper

contracts/src/River.1.sol
```diff
@@ -127,8 +135,20 @@ contract RiverV1 is
     }
 
     /// @inheritdoc IRiverV1
-    function initRiverV1_3(bytes32 withdrawalCredentails) external init(3) onlyAdmin {
-        initConsensusLayerDepositManagerV2(withdrawalCredentails);
+    function initRiverV1_3(
```
iamsahu, line 138: Why did we remove the `withdrawalCredentials` as a parameter & the call to `initConsensusLayerDepositManagerV2`?

contracts/src/River.1.sol
```diff
@@ -333,23 +381,70 @@ contract RiverV1 is
...
-            bytes[][] memory paddedPublicKeys = new bytes[][](operatorCount);
             for (uint256 i = 0; i < _fundedETH.length; ++i) {
                 paddedFundedETH[i] = _fundedETH[i];
-                uint256 pubKeyLength = _publicKeys[i].length;
-                if (pubKeyLength > 0) {
-                    paddedPublicKeys[i] = new bytes[](pubKeyLength);
-                    paddedPublicKeys[i] = _publicKeys[i];
-                } else {
-                    paddedPublicKeys[i] = new bytes[](0);
-                }
```
iamsahu, line 345: Why did we remove the padding code?

contracts/src/River.1.sol
```diff
@@ -127,8 +135,20 @@ contract RiverV1 is
...
+        address[] calldata _attesters,
+        uint256 _quorum,
+        bytes4 _genesisForkVersion
+    ) external init(3) onlyAdmin {
+        if (_depositDataBuffer == address(0)) revert LibErrors.InvalidZeroAddress();
+        if (_quorum == 0) revert ZeroQuorum();
+        if (_quorum > MAX_SIGNATURES) {
+            revert QuorumExceedsMaxSignatures(_quorum, MAX_SIGNATURES);
+        }
+        if (_attesters.length == 0) revert LibErrors.InvalidZeroAddress();
```
iamsahu: We should also enforce a check on the length so that it doesn't exceed the maximum amount.

contracts/src/libraries/BLS12_381.sol
iamsahu, line 1: We should copy over the tests from Lido into our repo for this library.

contracts/src/River.1.sol
juliaaschmidt: that's odd. Thanks great catch

contracts/src/libraries/BLS12_381.sol
juliaaschmidt: I was wondering whether I should do that too, but then checked their end-to-end test has a different license status. It is actually unlicensed I believe, which means we cannot freely copy it and use it for commercial purposes. cc @mischat

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -40,10 +64,46 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+    function _depositCommitteeQuorum() internal view override returns (uint256) {
+        return AttestationQuorum.get();
+    }
+
+    function _setDepositCommitteeQuorum(uint256 value) internal override {
+        AttestationQuorum.set(value);
+    }
+
+    function _depositDataBuffer() internal view override returns (IDepositDataBuffer) {
+        return IDepositDataBuffer(DepositDataBufferAddress.get());
```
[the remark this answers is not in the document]
juliaaschmidt: Reply to Copilot that the invariant is enforced at the boundary (init + admin setter), so a getter/validate() guard would be dead code. Validating at the entry point is preferable for two reasons:

  - Single source of truth: one rule to audit, not one per call site.
  - Cheaper: no SLOAD-then-compare on every deposit; the impossible state simply can't be reached.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,174 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        DepositDataBufferAddress.set(_depositDataBuffer);
+        emit SetDepositDataBuffer(_depositDataBuffer);
+    }
+
+    /// @notice Add or remove an attester. Admin only.
+    function setAttester(address attester, bool value) external onlyRiverAdmin {
+        if (attester == address(0)) revert ZeroAddress();
+
+        bool current = Attesters.isAttester(attester);
+        if (current == value) revert AttesterStatusUnchanged(attester, value);
```
iamsahu: No need to cache current.

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,312 @@
...
+    error InvalidSignatureLength(uint256 index, uint256 length);
+    error ZeroQuorum();
+    error ZeroDomainSeparator();
+    error ZeroDepositDomain();
+    error QuorumExceedsAttesterCount(uint256 quorum, uint256 attesterCount);
+    error QuorumExceedsMaxSignatures(uint256 quorum, uint256 max);
+    error TooManyAttesters(uint256 count, uint256 max);
+    error ZeroAddress();
+    error DuplicateAttester(address attester);
+    error AttesterStatusUnchanged(address attester, bool value);
```
iamsahu: Can we please add natspec for the errors.

juliaaschmidt: fixed

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -15,18 +17,40 @@ import "../state/river/CommittedBalance.sol";
...
+abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManagerV1, DepositToConsensusLayerValidation {
     /// @notice Size of a BLS Public key in bytes
     uint256 public constant PUBLIC_KEY_LENGTH = 48;
     /// @notice Size of a BLS Signature in bytes
     uint256 public constant SIGNATURE_LENGTH = 96;
     /// @notice Size of a deposit in ETH
     uint256 public constant DEPOSIT_SIZE = 32 ether;
 
+    /// @dev ASCII bytes for "operator:" prefix used in metadata encoding
+    bytes9 internal constant OPERATOR_PREFIX = "operator:";
```
mischat, line 39: Thanks trying to make this comply to the other deposit data buffer, but I feel like the business is going to want us to keep them seperate, to not co-mingle. 

we should ditch this generate metadata field, and add in a mandatory "nodeOperatorIndex" uint, which means we can loose the `_parseOperatorIndex` function that pulls out the data from the optional operator: string ... 

This will remove code, computation, and make it all cleaner IMO

mischat: when we make this change we will need to pull in the deposit databuffer into this codebase ... as it will be different from the one in the frontrun codebase

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -40,14 +64,50 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+
+    function _depositDomain() internal view override returns (bytes32) {
+        return DepositDomainValue.get();
+    }
+
     /// @notice Initializer to set the deposit contract address and the withdrawal credentials to use
     /// @param _depositContractAddress The address of the deposit contract
     /// @param _withdrawalCredentials The withdrawal credentials to apply to all deposits
-    function initConsensusLayerDepositManagerV1(address _depositContractAddress, bytes32 _withdrawalCredentials)
+    function initConsensusLayerDepositManagerV1_2(address _depositContractAddress, bytes32 _withdrawalCredentials)
```
mischat: do we even need t his, given that they would be set correctly in an older migration ?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+    /// @inheritdoc IConsensusLayerDepositManagerV1
+    function getAttesterCount() external view returns (uint256) {
+        return Attesters.getCount();
+    }
+
+    /// @inheritdoc IConsensusLayerDepositManagerV1
+    function getIsAttester(address attester) external view returns (bool) {
+        return Attesters.isAttester(attester);
+    }
+
```
mischat, line 169: should we call these something more specific, sorry i don't want to bikeshed ;) 

like getDepositDataAttestor (or however you spell it)

Given that we are going to have the consolidation attesters coming soon™

mischat: given that use : 

_depositCommitteeQuorum

how about 

_depositCommitteeAttester ?!?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+    /// @notice Add or remove an attester. Admin only.
+    function setAttester(address attester, bool status) external onlyRiverAdmin {
+        if (attester == address(0)) revert ZeroAddress();
+
+        if (Attesters.isAttester(attester) == status) revert AttesterStatusUnchanged(attester, status);
+
+        uint256 count = Attesters.getCount();
+        uint256 newCount = status ? count + 1 : count - 1;
```
mischat, line 189: i hate reading this notation ... i should get over it though ... and it correct anyways.

mischat, line 189: Can this underflow ... 

Where do we make sure that we don't have 0 attesters ?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        bytes32 depositDataBufferId,
+        bytes32 depositRootHash,
+        bytes[] calldata signatures,
+        BLS12_381.DepositY[] calldata depositYs
     ) external {
+        // 1. Keeper check
         if (msg.sender != KeeperAddress.get()) {
             revert OnlyKeeper();
         }
+
```
mischat, line 231: i guess we should put a comment on this one, we could move it up to check 0 ?!? or we re-number the rest of them

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,367 @@
...
+/// @notice Abstract contract that validates attestation quorum signatures and BLS deposit data.
+///         Storage hooks (_isAttester, _depositCommitteeQuorum) are virtual so that proxy-based deployments
+///         (River) can override them with unstructured storage.
+abstract contract DepositToConsensusLayerValidation {
+    // -----------------------------------------------------------------------
+    // Events
+    // -----------------------------------------------------------------------
+
+    /// @notice Emitted after attestation quorum is met and all deposits are executed.
+    event DepositsExecuted(bytes32 indexed depositDataBufferId, bytes32 indexed depositRootHash, uint256 depositCount);
```
mischat: DepositToConsensusLayerValidation.sol:20 declares event DepositsExecuted(bytes32 indexed depositDataBufferId, bytes32 indexed depositRootHash, uint256 depositCount). Nothing emits it. The actual emit is DepositsExecutedWithAttestation at ConsensusLayerDepositManager.1.sol:287. Either delete the declaration or wire it up — leaving an unused event in an abstract base will confuse indexers and event-typed integrations.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+            revert NotEnoughFunds();
         }
 
-        bytes32 withdrawalCredentials = WithdrawalCredentials.get();
-
-        if (withdrawalCredentials == 0) {
-            revert InvalidWithdrawalCredentials();
-        }
+        // 5. Update operator funded validator accounting
+        _updateFundedETHFromBuffer(deposits);
```
mischat, line 259: State updates happen AFTER external calls (CEI violation)

Lines 266 (_updateFundedETHFromBuffer → registry) and 270-274 (deposit{value: …} to the deposit contract) are external calls. Lines 277-285 (_setCommittedBalance, InFlightDeposit, TotalDepositedETH) happen after. Both callees are trusted protocol contracts so this isn't exploitable today, but if the registry or deposit-contract address gets repointed to malicious code by admin error or upgrade bug, the state would be stale during reentry. Move state writes ahead of the deposit loop (you can pre-compute the final values — no dependency on the loop succeeding because the whole tx is atomic).

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
-            publicKeys[i] = new bytes[](publicKeyCountPerOperator[i]);
-            // we reset the count to 0 so that we could reuse the array while adding the public keys in the loop below
-            publicKeyCountPerOperator[i] = 0;
         }
-
-        // Check if the total requested exceeds the committed balance
-        if (totalDeposits > committedBalance) {
-            revert ValidatorDepositsExceedCommittedBalance();
+        if (totalAmount > committedBalance) {
+            revert NotEnoughFunds();
```
mischat, line 255: They're the same kind of guard — "this batch fits in what we have" — just one is the accounting variable and one is the physical balance:

  ```// ConsensusLayerDepositManager.1.sol:261
  if (totalAmount > committedBalance) {
      revert NotEnoughFunds();
  }
  if (totalAmount > address(this).balance) {
      revert InsufficientBalance();          // new
  }```

  Why here:
  - totalAmount is already computed.
  - The reader sees both sides of the same invariant (accounting vs. physical) in one place.
  - Fails before any external call or state mutation, so no partial-state risk.

  Note: per-deposit, this is already implicitly caught — _depositValidator:353 does address(this).balance - _depositAmount, which under-flows (and reverts under 0.8 checked math) if the balance is short. So the aggregate check is purely fail-fast: revert on deposit 1 of N instead of deposit K of N with K external calls already burned.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        for (uint256 i = 0; i < len; i++) {
+            totalAmount += deposits[i].amount;
+            bytes32 depositWC;
+            bytes memory wc = deposits[i].withdrawalCredentials;
+            assembly {
+                depositWC := mload(add(wc, 32))
             }
-            if (_allocations[i].signature.length != SIGNATURE_LENGTH) {
-                revert InconsistentSignature();
+            if (depositWC != withdrawalCredentials) {
```
mischat: At validate() line 250, BLS verifies the signature against deposits[i].withdrawalCredentials. At lines 252-259 here, we then re-checks depositWC == withdrawalCredentials. Cleaner: pass the canonical River WC into validate() and use it as the WC for BLS verification directly. Then DepositObject.withdrawalCredentials becomes unused and can be dropped entirely. Eliminates an entire class of "what if the buffer producer lies about WC" reasoning.

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        // 5. Update operator funded validator accounting
+        _updateFundedETHFromBuffer(deposits);
 
+        // 6. Execute deposits
         address depositContract = DepositContractAddress.get();
-        uint256 operatorIndex;
-        for (uint256 idx = 0; idx < _allocations.length; ++idx) {
-            operatorIndex = _allocations[idx].operatorIndex;
+        for (uint256 i = 0; i < len; i++) {
             _depositValidator(
```
mischat, line 264: where do we check the depositRootHash against the one which the attesters attested to ? 

This is the AI output ... 

```
The depositDataRoot field is the bridge between attesters and the canonical deposit contract. The chain of trust is:

  1. Buffer producer publishes (pubkey, wc, sig, amount, depositDataRoot) per validator.
  2. Attesters independently verify each depositDataRoot against sha256 over (pubkey, wc, amount, sig), then sign (bufferId,
  depositRootHash).
  3. bufferId = keccak256(abi.encode(deposits)) binds the attestation to every field including depositDataRoot.
  4. On-chain, River verifies attester quorum, recomputes the bufferId, and passes the attested depositDataRoot to the deposit contract.
  5. The deposit contract is the verifier of last resort: it re-derives the root from the inputs and reverts if the attested value
  disagrees.

  The attesters' signature is what binds "this is the correct deposit data root for these inputs"; the deposit contract is what enforces
  it. River's role is to faithfully forward the attested value, not to compute its own.

  What the current code does wrong

  _depositValidator:346-351 recomputes the root locally and :356 passes the local value. This severs the attester→deposit-contract bridge:

  - Attesters could sign deposits with a garbage depositDataRoot and the tx would still succeed (River silently overrides them with the
  correct local value).
  - Conversely, the deposit-contract revert that should fire when attesters are wrong never gets exercised, so the attester verification
  step is unverifiable from on-chain behavior — you lose a defense-in-depth check that the design was buying you.

  The check exists in the deposit contract; River just isn't using it.

  The fix

  Drop the local derivation; pass the buffer's value through.

  // ConsensusLayerDepositManager.1.sol — replace lines 338-351 with: nothing
  // and change the call site to use the attested root.

  function _depositValidator(
      bytes memory _publicKey,
      bytes memory _signature,
      uint256 _depositAmount,
      bytes32 _withdrawalCredentials,
      bytes32 _depositDataRoot,         // ← new param
      address _depositContract
  ) internal {
      if (_depositAmount < 1 ether || _depositAmount > 2048 ether || _depositAmount % 1 gwei != 0) {
          revert InvalidDepositSize(_depositAmount);
      }

      uint256 targetBalance = address(this).balance - _depositAmount;
      IDepositContract(_depositContract).deposit{value: _depositAmount}(
          _publicKey, abi.encodePacked(_withdrawalCredentials), _signature, _depositDataRoot
      );
      if (address(this).balance != targetBalance) revert ErrorOnDeposit();
  }

  And at the caller (depositToConsensusLayerWithAttestation:271):

  _depositValidator(
      deposits[i].pubkey,
      deposits[i].signature,
      deposits[i].amount,
      withdrawalCredentials,
      deposits[i].depositDataRoot,   // ← attested value, now actually used
      depositContract
  );

  This:
  - Restores the design — the attested depositDataRoot is what flows to the deposit contract, so the deposit contract becomes the on-chain
  enforcement of "did attesters compute it correctly?"
  - Saves ~30k gas per deposit (three sha256 calls + memory shuffling deleted).
  - Removes any reason for depositDataRoot to feel decorative in the struct.

```

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,367 @@
...
+                    break;
+                }
+            }
+            if (duplicate) continue;
+
+            seen[validCount] = signer;
+            validCount++;
+        }
+
+        uint256 depositCommitteeQuorum = _depositCommitteeQuorum();
```
mischat: _depositCommitteeQuorum is called multiple times in this method, we can use a variable ...

contracts/src/components/ConsensusLayerDepositManager.1.sol
mischat: but i believe the AI to be wrong here ... it looks like the root hash is validated against the onchain one in, literally in the first test in the validate function !!!

`_verifyAttestationQuorum` 

on line 207 in contracts/src/components/DepositToConsensusLayerValidation.sol

bad hallucinating AI ...

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+        _setDepositCommitteeQuorum(newQuorum);
+        emit SetAttestationQuorum(newQuorum);
+    }
+
+    // -----------------------------------------------------------------------
+    // Attestation-based deposit function
+    // -----------------------------------------------------------------------
+
+    /// @inheritdoc IConsensusLayerDepositManagerV1
+    function depositToConsensusLayerWithAttestation(
```
mischat, line 221: shall we do non reentrant ?

It calls out to different contracts albeit it our contracts and the official deposit contract ... just incase, as this is the most sensitive thing we do ?

contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -95,93 +148,173 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
...
+
+    // -----------------------------------------------------------------------
+    // Internal — metadata parsing
+    // -----------------------------------------------------------------------
+
+    /// @notice Parse an operator index from a bytes32 metadata field.
+    ///         Expected format: left-aligned ASCII "operator:N" zero-padded on the right.
+    /// @param metadata The metadata bytes32 value
+    /// @return operatorIndex The parsed operator index
+    function _parseOperatorIndex(bytes32 metadata) internal pure returns (uint256 operatorIndex) {
```
mischat, line 291: as per above, in a future PR, i think we should kill this and have a customised nodeOperatorId field in this struct ...

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,367 @@
...
+    error AttesterStatusUnchanged(address attester, bool status);
+
+    // -----------------------------------------------------------------------
+    // EIP-712
+    // -----------------------------------------------------------------------
+
+    bytes32 internal constant EIP712_DOMAIN_TYPEHASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+    bytes32 internal constant NAME_HASH = keccak256("DepositToConsensusLayerValidation");
+    bytes32 internal constant VERSION_HASH = keccak256("1");
```
mischat, line 96: what is this, the version of the validation function, should we make it 1.0.0, not sure what this number should be ?

contracts/src/components/DepositToConsensusLayerValidation.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    /// @notice Validate attestation signatures and BLS deposit data.
+    ///         Returns the validated deposits array for use by the caller.
+    /// @param depositDataBufferId  Batch identifier in the DepositDataBuffer
+    /// @param depositRootHash      Current deposit contract root hash co-signed by attesters
+    /// @param signatures           EIP-712 signatures from attesters
+    /// @param depositYs            Y-coordinates for BLS decompression, one per deposit
+    /// @return deposits            The validated deposit batch
+    // solhint-disable-next-line code-complexity
+    function validate(
```
mischat, line 190: Perhaps we can ignore this, but worth flagging here : 

``` 
Five loops over deposits per call

  validate() does length checks + BLS verification (2 passes). _updateFundedETHFromBuffer does parse-indices + sum/bucket (2 passes). Then this function does WC + totalAmount (1 pass) and the deposit loop (1 pass). The WC + totalAmount loop could be fused into validate() since both depend on data already loaded there. Probably not worth the readability hit unless gas matters.```

File: contracts/src/components/DepositToConsensusLayerValidation.sol [hunk 43]
```diff
@@ -0,0 +1,367 @@
...
+    // -----------------------------------------------------------------------
+
+    /// @notice Validate attestation signatures and BLS deposit data.
+    ///         Returns the validated deposits array for use by the caller.
+    /// @param depositDataBufferId  Batch identifier in the DepositDataBuffer
+    /// @param depositRootHash      Current deposit contract root hash co-signed by attesters
+    /// @param signatures           EIP-712 signatures from attesters
+    /// @param depositYs            Y-coordinates for BLS decompression, one per deposit
+    /// @return deposits            The validated deposit batch
+    // solhint-disable-next-line code-complexity
```
mischat: i have taken notes on how is compares to the one in the frontrun mitigation, my aim would be to combine these in a library so that we don't have duplicate code anywhere.

contracts/src/interfaces/components/IConsensusLayerDepositManager.1.sol
```diff
@@ -1,9 +1,10 @@
 //SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.34;
 
-import "../IOperatorRegistry.1.sol";
+import "../IDepositDataBuffer.sol";
+import "../../libraries/BLS12_381.sol";
 
-/// @title Consensys Layer Deposit Manager Interface (v1)
```
mischat, line 6: hehe

contracts/src/interfaces/IDepositDataBuffer.sol
```diff
@@ -0,0 +1,72 @@
...
+        /// @dev Deposit amount in wei (must be a multiple of 1 gwei and typically 32 ether)
+        uint256 amount;
+        /// @dev 32-byte withdrawal credentials (abi-encoded bytes32)
+        bytes withdrawalCredentials;
+        /// @dev SHA-256 deposit data root for use with the official deposit contract
+        bytes32 depositDataRoot;
+        /// @dev Arbitrary metadata bytes32. River encodes the operator index as left-aligned ASCII:
+        ///      "operator:N" zero-padded on the right. E.g. "operator:42" →
+        ///      0x6f70657261746f723a3432000000000000000000000000000000000000000000
+        bytes32 metadata;
```
mischat, line 24: i believe we should make this more specific to LC ... as per previous comment.

contracts/src/interfaces/IRiver.1.sol
```diff
@@ -161,8 +161,22 @@ interface IRiverV1 is IConsensusLayerDepositManagerV1, IUserDepositManagerV1, IS
...
+    ///         quorum, buffer, domain) in a single admin-gated transaction so there is no race window
+    ///         on the attestation trust anchor.
+    /// @param _withdrawalCredentials The withdrawal credentials to apply to all deposits
+    /// @param _depositDataBuffer The address of the deposit data buffer
+    /// @param _attesters The list of attesters to add
+    /// @param _quorum The attestation quorum
+    /// @param _genesisForkVersion The genesis fork version for computing the deposit domain
+    function initRiverV1_3(
+        bytes32 _withdrawalCredentials,
+        address _depositDataBuffer,
```
mischat, line 175: given that we will make this specific to LC, it is worth having the DepositDataBuffer be called the DepositManager or something, and then we can try and define a shared IDepositDataBuffer !?! just a thought for the future.

contracts/src/libraries/BLS12_381.sol
mischat: i have a load of tests in the frontrun, we can figure this out when we decide where the code should live.

contracts/src/state/river/AttestationQuorum.sol
```diff
@@ -0,0 +1,23 @@
+//SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+
+/// @title AttestationQuorum
+/// @notice Library for storing the attestation quorum.
+library AttestationQuorum {
+    bytes32 internal constant ATTESTATION_QUORUM_SLOT =
+        bytes32(uint256(keccak256("river.state.attestationQuorum")) - 1);
```
mischat, line 10: when we decide where the code lives we should move all of their namespaces ...

mischat (dismissed):
I have left a bunch of comments, mainly so that we can open issues off the back of them. 

LGTM

contracts/src/state/river/AttestationQuorum.sol
juliaaschmidt: I have moved quorum-related state in this PR: https://github.com/liquid-collective/liquid-collective-protocol/pull/488/changes

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: Great catch!

contracts/src/components/ConsensusLayerDepositManager.1.sol
iamsahu: We don't need to add this check as it is an unreachable condition. However, if we are able to reach this condition it would mean that our accounting is broken as committedBalance is set incorrectly.

This check just makes the transaction fail early if our accounting was broken. So, in my opinion we should not add this check.

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: If depositToConsensusLayer is onlyKeeper do we need this? cc @iamsahu

contracts/src/components/DepositToConsensusLayerValidation.sol
juliaaschmidt: This is the same as [this](https://github.com/AlluvialFinance/frontrun-mitigation/blob/201697e0104a45a69b399e030bce87f19e63a1d1/src/DepositToConsensusLayerValidation.sol#L88). Happy to refactor it as it is in the front-run repo if that is more legible?

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: I'll address this on top of my size-based refactor PR. The issue is tracked here https://github.com/liquid-collective/liquid-collective-protocol/issues/494

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: I kept it like this thinking it would be easily modifiable if we ever needed to do TVS transfers and support multiple WC. I have made your suggested change now. This simplifies the code a lot.

contracts/src/components/DepositToConsensusLayerValidation.sol
juliaaschmidt: fixed here https://github.com/liquid-collective/liquid-collective-protocol/pull/396/commits/cf7992fdb88840dcf0992636b44761e326e10ea0

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: fixed here https://github.com/liquid-collective/liquid-collective-protocol/pull/396/commits/8213325f0e72b22e0ce115c33679875d7c63760e

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: No, I added a comment to why https://github.com/liquid-collective/liquid-collective-protocol/pull/396/commits/1377b32b69ef7495f1003238dc76350cdc09380c

contracts/src/components/ConsensusLayerDepositManager.1.sol
juliaaschmidt: I have made the change

contracts/src/components/DepositToConsensusLayerValidation.sol
juliaaschmidt: This can be attempted post-refactor PR. Hence this is tracked in a separate issue https://github.com/liquid-collective/liquid-collective-protocol/issues/497

iamsahu (approved):
LGTM

contracts/src/components/ConsensusLayerDepositManager.1.sol
mischat: thanks, happy to close this.

contracts/src/components/ConsensusLayerDepositManager.1.sol
mischat: we don't need it now, it is just incase we upgrade one of the contracts it speaks too, etc etc ... but yeah, probably fine to leave it

contracts/src/components/DepositToConsensusLayerValidation.sol
mischat: ok ... thanks ... i can;t recall that, but it seems to make sense

mischat (approved):
LGTM !

PR #402, opened 2026-04-02
Merge main into pectra branch post BYOV audit (BS-3198)

iamsahu:
## Description

Merged main into v2/pectra branch after BYOV audit fixes were signed off & merged into main.

## Notice

## Pull Request Type

## Testing


iamsahu:
@copilot resolve the merge conflicts in this pull request

juliaaschmidt (approved):
LGTM

contracts/src/River.1.sol
```diff
@@ -619,7 +649,11 @@ contract RiverV1 is
     /// @notice Committed funds are funds waiting to be deposited but that cannot be used to fund the redeem manager anymore
     /// @notice This two step process is required to prevent possible out of gas issues we would have from actually funding the validators at this point
     /// @param _period The period between current and last report
-    function _commitBalanceToDeposit(uint256 _period) internal override {
+    function _commitBalanceToDeposit(uint256 _period, bool _slashingContainmentModeEnabled) internal override {
+        if (_slashingContainmentModeEnabled) {
+            return;
+        }
```
mischat, line 655: Do we want this to just return here ? do we not want to at least emit an event ?

contracts/src/interfaces/IRedeemManager.1.sol
```diff
@@ -158,6 +161,15 @@ interface IRedeemManagerV1 {
...
 
+    /// @notice Creates a redeem request
+    /// @param _lsETHAmount The amount of LsETH to redeem
+    /// @param _recipient The recipient owning the redeem request
+    /// @param _initiator The initiator of the request
+    /// @return redeemRequestId The id of the redeem request
+    function requestRedeem(uint256 _lsETHAmount, address _recipient, address _initiator)
+        external
+        returns (uint32 redeemRequestId);
+
```
mischat, line 172: was this just not there before ?

contracts/src/River.1.sol
mischat: This is claude's answer : 

```
⏺ Answer

  Caller. _commitBalanceToDeposit is a virtual hook declared in contracts/src/components/OracleManager.1.sol:91 and implemented at
  contracts/src/River.1.sol:652. There's exactly one caller: OracleManager.1.sol:462, inside setConsensusLayerData (the oracle report
  ingestion path).

  Empty return handling. There's nothing to handle — the function is void. It only mutates two storage slots (CommittedBalance,
  BalanceToDeposit). Slashing containment just skips those mutations. The caller continues with emit ProcessedConsensusLayerReport(_report,
   vars.trace) at OracleManager.1.sol:465.

  Why not revert? Reverting would abort the entire oracle report — pulling CL funds, rewards accounting, _reportCLETH, exit requests,
  redeem-manager withdraw reporting, etc. The intent of slashing containment is the opposite: keep processing the report normally but pause
   the deposit-side firehose (no new 32-ETH chunks moved into CommittedBalance) while a slashing event is being investigated. Same pattern
  is used in the sister hook at River.1.sol:587-589 for _requestExitsBasedOnRedeemDemandAfterRebalancings.

  Why no event? The flag is already observable on-chain in two ways:
  1. storedReport.slashingContainmentMode = _report.slashingContainmentMode at OracleManager.1.sol:352 — persisted in
  LastConsensusLayerReport.
  2. The full _report is emitted in ProcessedConsensusLayerReport (OracleManager.1.sol:465), which carries slashingContainmentMode.

  So a consumer can already detect "this report ran with containment on" without a dedicated event.

  That said — a fair critique. A targeted event like DepositCommittalSkippedDueToSlashingContainment(period) (and a matching one in the
  exits path) would be cheap and materially better for monitoring/alerting than indexing the whole report struct. It's also the kind of
  "silent skip" that tends to be missed in dashboards. Worth raising as a PR comment — it's an observability nit, not a correctness bug.
```

contracts/src/River.1.sol
```diff
@@ -222,10 +222,17 @@ contract RiverV1 is
     }
 
     /// @inheritdoc IRiverV1
-    function requestRedeem(uint256 _lsETHAmount, address _recipient) external returns (uint32 _redeemRequestId) {
+    function requestRedeem(uint256 _lsETHAmount, address _recipient)
+        external
+        whenNotSlashingContainmentMode
```
mischat, line 227:
```
⏺ Is the guard at the right level?

  Deposit side: yes, the placement is correct.

  deposit() / depositAndTransfer() / receive()  →  _deposit()  →  _setBalanceToDeposit + _onDeposit
                                                      ↑
                                                    guard

  The guard sits at River._deposit (River.1.sol:484) — the deepest single chokepoint reached by all three external entry points, executed
  before any storage mutation, in the same contract that owns the slashing-containment state. One modifier, three callsites covered, no
  cross-contract call. That's textbook chokepoint placement.

  Redeem side: functionally correct, but architecturally fragile. Three concerns:

  1. The guard is replicated, not consolidated

  There are now two whenNotSlashingContainmentMode modifiers — one in River.1.sol:476 (reads its own storage), one in
  RedeemManager.1.sol:56 (does a cross-contract STATICCALL to River). Five guarded externals across two contracts:

  River.requestRedeem(uint, addr)              ← guarded (River-local read)
  RedeemManager.requestRedeem(uint, addr, addr) ← UNguarded (relies on caller)
  RedeemManager.requestRedeem(uint, addr)       ← guarded (STATICCALL to River)
  RedeemManager.requestRedeem(uint)             ← guarded (STATICCALL to River)
                         ↓
                  _requestRedeem()  ← UNguarded internal sink

  The deposit side has one chokepoint (_deposit). The redeem side has three chokepoints, all upstream of the actual sink (_requestRedeem).
  That's the asymmetry.

  2. The 3-arg overload is only transitively protected

  RedeemManager.requestRedeem(uint, addr, addr) (:154) has only onlyRiver — no slashing modifier. It relies on the implicit invariant:
  every River-side caller of this function also has whenNotSlashingContainmentMode. Today there's exactly one River caller
  (River.requestRedeem) and it does. But that's a invariant-by-convention, not enforced by the type system. A future River change adding a
  second call path could silently bypass the guard.

  3. The cross-contract modifier has an init-order footgun

  RedeemManager's modifier does _castedRiver().getSlashingContainmentMode() — a STATICCALL. The natspec at RedeemManager.1.sol:54-55
  literally warns: "River must be initialized before any function guarded by this modifier is callable; calls will revert with a
  non-contract error otherwise." That's a deployment/upgrade-ordering hazard introduced by guarding at the external layer instead of at
  _requestRedeem.

  Recommendation

  Consolidate to the deposit-side pattern: move the guard to RedeemManager._requestRedeem (the single internal sink), then drop the three
  external modifiers. That gives you:

  - One chokepoint, same as _deposit
  - The 3-arg overload becomes implicitly safe — no convention to maintain
  - One STATICCALL per request (same as today; you can't avoid the cross-contract read since the state lives on River)
  - River's external requestRedeem keeps its own River-local modifier as an early-revert (saves the LSETH _transfer gas before STATICCALL)

  If you don't want to refactor, the minimal hardening on the current PR is: add whenNotSlashingContainmentMode to the 3-arg overload too,
  so the implicit "every River caller must remember" invariant becomes explicit.

  Adjacent question worth raising on the PR

  ETH-receiving externals (sendCLFunds, sendELFees, sendCoverageFunds, sendRedeemManagerExceedingFunds) are all unguarded. That's almost
  certainly intentional — containment is meant to pause new validator funding and shareholder churn, not freeze the reward-pull pipeline —
  but the spec should explicitly say so. The PR description is the right place to document "containment intentionally does not block reward
   inflows; the brake is applied downstream at _commitBalanceToDeposit and _requestExitsBasedOnRedeemDemandAfterRebalancings."
```

contracts/src/River.1.sol
```diff
@@ -455,6 +462,29 @@ contract RiverV1 is
         BalanceToRedeem.set(_newBalanceToRedeem);
     }
 
+    /// @notice Returns whether slashing containment mode is currently active
+    function _getSlashingContainmentMode() internal view override(ConsensusLayerDepositManagerV1) returns (bool) {
```
mischat, line 466: containment is meant to pause new validator funding and shareholder churn, not freeze the reward-pull pipeline — but the spec should explicitly say so. The PR description is the right place to document "containment intentionally does not block reward inflows; the brake is applied downstream at _commitBalanceToDeposit and _requestExitsBasedOnRedeemDemandAfterRebalancings."

contracts/src/RedeemManager.1.sol
```diff
@@ -41,20 +42,20 @@ contract RedeemManagerV1 is Initializable, IRedeemManagerV1, IProtocolVersion {
         _;
     }
 
-    modifier onlyRedeemerOrRiver() {
```
mischat, line 44: was this never used before ? or was it only used by the OperatorsRegistry and that stuff that we have removed ?

contracts/src/RedeemManager.1.sol
```diff
@@ -149,22 +150,37 @@ contract RedeemManagerV1 is Initializable, IRedeemManagerV1, IProtocolVersion {
         }
     }
 
+    /// @inheritdoc IRedeemManagerV1
+    function requestRedeem(uint256 _lsETHAmount, address _recipient, address _initiator)
+        external
+        onlyRiver
```
mischat, line 156: should this also have a `whenNotSlashingContainmentMode` modifier ?

contracts/src/RedeemManager.1.sol
```diff
@@ -312,8 +329,12 @@ contract RedeemManagerV1 is Initializable, IRedeemManagerV1, IProtocolVersion {
     /// @notice Perform a new redeem request for the specified recipient
     /// @param _lsETHAmount The amount of LsETH to redeem
     /// @param _recipient The recipient owning the request
+    /// @param _initiator The initiator of the request
     /// @return redeemRequestId The id of the newly created redeem request
-    function _requestRedeem(uint256 _lsETHAmount, address _recipient) internal returns (uint32 redeemRequestId) {
+    function _requestRedeem(uint256 _lsETHAmount, address _recipient, address _initiator)
+        internal
```
mischat, line 335: could we just have the whenNotslashcontainmentMode here ?

contracts/src/RedeemManager.1.sol
mischat: posting this : 

```
Redeem side: functionally correct, but architecturally fragile. Three concerns:

  1. The guard is replicated, not consolidated

  There are now two whenNotSlashingContainmentMode modifiers — one in River.1.sol:476 (reads its own storage), one in
  RedeemManager.1.sol:56 (does a cross-contract STATICCALL to River). Five guarded externals across two contracts:

  River.requestRedeem(uint, addr)              ← guarded (River-local read)
  RedeemManager.requestRedeem(uint, addr, addr) ← UNguarded (relies on caller)
  RedeemManager.requestRedeem(uint, addr)       ← guarded (STATICCALL to River)
  RedeemManager.requestRedeem(uint)             ← guarded (STATICCALL to River)
                         ↓
                  _requestRedeem()  ← UNguarded internal sink

  The deposit side has one chokepoint (_deposit). The redeem side has three chokepoints, all upstream of the actual sink (_requestRedeem).
  That's the asymmetry.

  2. The 3-arg overload is only transitively protected

  RedeemManager.requestRedeem(uint, addr, addr) (:154) has only onlyRiver — no slashing modifier. It relies on the implicit invariant:
  every River-side caller of this function also has whenNotSlashingContainmentMode. Today there's exactly one River caller
  (River.requestRedeem) and it does. But that's a invariant-by-convention, not enforced by the type system. A future River change adding a
  second call path could silently bypass the guard.

  3. The cross-contract modifier has an init-order footgun

  RedeemManager's modifier does _castedRiver().getSlashingContainmentMode() — a STATICCALL. The natspec at RedeemManager.1.sol:54-55
  literally warns: "River must be initialized before any function guarded by this modifier is callable; calls will revert with a
  non-contract error otherwise." That's a deployment/upgrade-ordering hazard introduced by guarding at the external layer instead of at
  _requestRedeem.

  Recommendation

  Consolidate to the deposit-side pattern: move the guard to RedeemManager._requestRedeem (the single internal sink), then drop the three
  external modifiers. That gives you:

  - One chokepoint, same as _deposit
  - The 3-arg overload becomes implicitly safe — no convention to maintain
  - One STATICCALL per request (same as today; you can't avoid the cross-contract read since the state lives on River)
  - River's external requestRedeem keeps its own River-local modifier as an early-revert (saves the LSETH _transfer gas before STATICCALL)

  If you don't want to refactor, the minimal hardening on the current PR is: add whenNotSlashingContainmentMode to the 3-arg overload too,
  so the implicit "every River caller must remember" invariant becomes explicit.
```

contracts/src/River.1.sol
mischat: cross-posting the same comment.

mischat (approved):
I would like to go through my comments, so that we can create issues out of them or resolve them before approving.

contracts/src/interfaces/IRedeemManager.1.sol
iamsahu: This wasn't there before, we had to create a new method so that we could track the initiator correctly.

contracts/src/RedeemManager.1.sol
iamsahu: It was used before as there was only a single requestRedeem function. Now we have 2 variants of it that's why it is no longer required.

contracts/src/RedeemManager.1.sol
iamsahu: We do have the modifier in the river contract from where this gets called.

contracts/src/RedeemManager.1.sol
iamsahu: Yeah, we could have done that, but we wanted to preserve order of failure so we choose to keep it at the external function.

contracts/src/River.1.sol
iamsahu: Created an issue for this.

contracts/src/River.1.sol
iamsahu: Captured this in an issue

deploy/devHoodi/03_deploy_river_oracle_operators_registry_and_redeem_manager.ts
```diff
@@ -16,16 +16,100 @@ import { verify } from "../../scripts/helpers";
...
 const func: DeployFunction = async function ({
   deployments,
   getNamedAccounts,
   ethers,
   network,
 }: HardhatRuntimeEnvironment) {
-  if (!["hardhat", "local", "tenderly", "hoodi", "devHoodi", "kurtosis"].includes(network.name)) {
+  if (!["hardhat", "local", "tenderly", "hoodi", "devHoodi"].includes(network.name)) {
     throw new Error("Invalid network for hoodi deployment");
   }
```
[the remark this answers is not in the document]
mischat: this isn't relevant

PR #471, opened 2026-05-05
test: add autocompounding support and exit-with-rewards test

### LINKED ISSUE #433: Incorrect test conditions
How does this test pass at the moment, when you only allow to exit as much as was previously deposited but not deposit+rewards? I guess it works because the rewards are too tiny to yield measurable balance increases (autocompounding)

_Originally posted by @juliaaschmidt in https://github.com/liquid-collective/liquid-collective-protocol/pull/395#discussion_r3087565009_

iamsahu:
## Summary
- Adds `sim_autocompound()` to `BeaconChainSimulator` to model Pectra (0x02) autocompounding where CL rewards increase validator balance instead of being skimmed
- Tracks cumulative autocompounded rewards via `_simCumulativeAutocompounded`
- Updates I2 invariant upper bound to include autocompounded rewards
- Adds `testExitWithAutocompoundedRewards` to verify exits can return more ETH than the original deposit when rewards have autocompounded

Previously exit tests only passed because `sim_advanceEpoch` modeled all rewards as skimming (0x01 behavior), keeping validator `currentBalance` constant. With Pectra, this was an incorrect test condition.

Closes #433

## Test plan
- [x] `forge test --match-path "contracts/test/accounting/*"` — 40 tests pass (including new test)
- [x] Fuzz suite passes with updated I2 invariant


contracts/test/accounting/BeaconChainSimulator.sol
```diff
@@ -94,14 +97,25 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
     function sim_advanceEpoch(uint256 rewardsPerValidator) internal {
         for (uint256 i = 0; i < _simValidators.length; i++) {
             if (_simValidators[i].state == ValidatorState.Active) {
-                // Rewards are swept (skimmed) from the CL to EL each epoch.
-                // The validator's CL balance remains at the principal (DEPOSIT_SIZE)
-                // after the sweep, so we only track cumulative skimmed rewards separately.
+                // Models 0x01 (BLS) withdrawal credentials: rewards are swept (skimmed)
```
juliaaschmidt: Naming nit: sim_advanceEpoch reads like it should advance epoch state, but it doesn't; it just credits one epoch's worth of skimmed rewards (assuming 0x01 validator). The current name overstates what it does.

 It's also worth noting that the simulated validators don't carry a WC-kind field, so the choice between 0x01 (skimming) and 0x02 (compounding) is made entirely at the call site by which sibling function the test invokes. So you could treat a validator as 0x01 and 0x02 in the same test. 

  Two suggestions:
  1. Rename to `sim_skimRewards` (or sim_creditSkimmedRewards), symmetric with `sim_autocompound` and accurate about behavior.
  2. Add a one-liner on both functions explicitly noting the mode is selected per call, not per validator, so the credential model assumption is hard to miss when reading a test.

juliaaschmidt (dismissed):

mischat:
@coderabbitai review

juliaaschmidt (approved):

mischat (approved):
MERGE it

PR #472, opened 2026-05-05
test: add unit tests for reportCLETH and demandETHExits

### LINKED ISSUE #429: More tests for new oracle values.
```
 ---
  What's NOT well tested

  reportCLETH — this function has no direct unit test at all. It is only exercised indirectly through River integration tests when an oracle report
  flows through. There is no test for:
  - The InvalidActiveCLETHArrayLength revert (array length ≠ operator count)
  - The InvalidEmptyArray revert (zero-length array)
  - That operator.activeCLETH is actually set correctly per-operator
  - The UpdatedActiveCLETH event emission

  incrementFundedETH — the OperatorIgnoredExitRequests check against exitedETH — the existing test
  (testIncrementFundedRevertsOperatorIgnoredExitRequests) uses sudoSetFunded and sudoExitRequests with integer counts (not ETH), meaning it exercises
  the check but not with a live exitedETH array that the check actually reads from. The branch where the operator has partially fulfilled exits
  (exitedETH > 0 but < requestedExits) is not tested through the real funding path end-to-end.

  demandETHExits — the cap logic — the availableCLETHAfterCurrentETHExitsDemand calculation is tested in scenarios where the new demand fits, but the
  cap path (where _exitAmountToRequest > availableCLETH - currentDemand and is silently reduced) has only a couple of unit tests. There is no test
  where _totalAvailableCLETH < currentETHExitsDemand (the ? 0 branch makes available = 0, silently swallowing the request).

  The commented-out test testDepositThenExitEndToEnd (line 1066) — the full cycle of deposit → exit request → exits fulfilled → re-deposit for the same
   operator was never rewritten after the old key-based deposit path was removed. This is the most important end-to-end integration scenario missing.

```

And re: invariant tests; 

```
 Not covered by any invariant:
  - CurrentETHExitsDemand and TotalETHExitsRequested are never touched by any invariant — the sum demand + requested relative to TotalDepositedETH or
  totalAvailableCLETH is never asserted
  - operator.activeCLETH values are never invariant-checked
  - The reportCLETH write is invisible to the invariant suite entirely
```

_Originally posted by @mischat in https://github.com/liquid-collective/liquid-collective-protocol/pull/395#discussion_r3137150598_

iamsahu:
## Summary
- Adds 4 direct unit tests for `reportCLETH`: empty array revert, wrong length revert, per-operator value set, `UpdatedActiveCLETH` event emission
- Adds 3 unit tests for `demandETHExits`: cap path when request exceeds available CLETH, zero available swallows request, accumulated demand blocks further demand
- These functions had no direct unit tests previously — only exercised indirectly through integration paths

Closes #429

## Test plan
- [x] `forge test --match-path "contracts/test/OperatorsRegistry.1.t.sol"` — 69 tests pass


juliaaschmidt (approved):
🔥 Thanks for adding these

mischat (approved):
🚢

PR #474, opened 2026-05-05
test: add tests for fundedETH padding branch in River

### LINKED ISSUE #426: Incomplete tests for padding
This section isn't tested well enough IMO ... 

```
  What is NOT tested:

  The padding branch — the code path at River.1.sol:327-340 that fires when _fundedETH.length < operatorCount. This path triggers when the deposit
  batch only touches some operators, meaning the array passed from ConsensusLayerDepositManager is shorter than the total operator count in the
  registry. None of the tests deposit to a strict subset of operators where that subset count is less than the total operator count registered in River
   itself (they test via the deposit manager harness, not through River end-to-end with a larger operator set). So the padding + public key copying
  loop has no direct test coverage.
```

_Originally posted by @mischat in https://github.com/liquid-collective/liquid-collective-protocol/pull/395#discussion_r3137095151_

iamsahu:
## Summary
- Adds `FundedETHPadding.t.sol` with 3 tests covering the padding branch in `River._incrementFundedETH` (line 334)
- **Subset deposit**: deposits only to operator 0 when 2 operators exist, triggering the padding path
- **Sequential deposits**: deposits to operator 0 then operator 1, verifying both funded values are correct
- **Full exit cycle**: deposits with padding, then exits, verifying padding doesn't corrupt funded/exited tracking
- All invariants (I1-I6) checked after each oracle report

Closes #426

## Test plan
- [x] `forge test --match-path "contracts/test/accounting/scenarios/FundedETHPadding*"` — 3 tests pass
- [x] Full accounting suite passes (43 tests)


juliaaschmidt (commented):
Given the changes requested in #489 that are being addressed in PR #488, I don't think the PR as it is makes sense anymore. We should update it to test the new OperatorFundingDelta branch 🙏

juliaaschmidt (changes requested):
I think this PR is no longer required since the padding has been removed as per https://github.com/liquid-collective/liquid-collective-protocol/issues/489

mischat:
@coderabbitai review

mischat:
there is no more padding logic ... this has moved to the LibFundingDelta

PR #476, opened 2026-05-05
test: add missing invariants I7-I10 and deposit share price check

### LINKED ISSUE #417: More invariants to be added for testing
I think there is some good stuff in here : 

but looking at the design, the key invariant here is around the inFlightEth and the accounting around that in the Oracle report : 

```  ---
  Missing invariants — what I'd add

  Missing 1 — The central accounting identity is not checked

  The most important invariant in the whole new design is not in here:

  InFlightDeposit + storedReport.totalDepositedActivatedETH == TotalDepositedETH

  Every ETH ever sent to the deposit contract is either still in-flight or has activated. These two buckets should sum to exactly
  TotalDepositedETH. I9 only checks InFlightDeposit <= TotalDepositedETH. The exact decomposition — which is what makes _assetBalance()
  correct — has no invariant. If a bug caused these three values to diverge, the share price would silently drift. This should be tested
  continuously.

  Missing 2 — requestedExits >= exited per operator

  I18 checks exited <= funded and requestedExits <= funded, but not the specific ordering between them: requestedExits >= exited. This
  ordering is load-bearing — requestETHExits computes opPendingExits = requestedExits - getExitedETHAtIndex(idx) and relies on it being
  non-negative. _setExitedETH is supposed to maintain this by bumping requestedExits whenever exitedETH overtakes it, but there's no
  invariant verifying that guarantee holds continuously.

  Missing 3 — TotalETHExitsRequested >= totalExited

  Every time an exit is processed, either it was explicitly requested (which incremented TotalETHExitsRequested) or it was unsolicited
  (which also increments TotalETHExitsRequested via the unsolicited path in _setExitedETH). So TotalETHExitsRequested should always be >=
  totalExited. Not currently checked.

  Missing 4 — Share price stable across deposits

  I1 only fires after oracle reports. A deposit should also not move the share price (that's the whole point of InFlightDeposit). There's
  no continuous check that _assetBalance / totalShares stays flat when depositToConsensusLayerWithDepositRoot is called. The simulator
  checks the InFlightDeposit value immediately after deposit (inline assert in sim_deposit) but doesn't verify the share price is
  unaffected.

  Missing 5 — CurrentETHExitsDemand + TotalETHExitsRequested <= TotalDepositedETH

  The total amount the protocol has either requested or is still demanding in exits should be bounded by what was ever deposited.
  Compounding in Pectra means this bound is slightly more complex (activated validators can grow above their original deposit), but at
  minimum it should hold against TotalDepositedETH as a coarse check.

  Missing 6 — totalDepositedActivatedETH non-decreasing (continuous, not just per-report)

  I15/I16 use ghost variables to check skimmed and exited don't decrease across reports. The same treatment should be applied to
  totalDepositedActivatedETH. Currently there's no ghost tracking it between calls — it's only validated inside setConsensusLayerData via
  the revert path. The invariant test should enforce it from the outside too.

  ---
  Other observations on the test setup

  A few things worth noting about coverage gaps in the handler itself, independent of the invariants:

  - Only 2 operators are ever used. Invariants involving per-operator arrays are only tested with 2 entries. Adding a third operator
  mid-test (or testing the case where the exitedETH array hasn't grown to cover a new operator) is never exercised.
  - I13 was intentionally dropped because the harness bypasses 32-ETH alignment. That's fine, but it means CommittedBalance % 32 ETH == 0
  is never validated in fuzzing.
  - EL fee pulls are not simulated. sim_advanceEpoch only adds to _simCumulativeSkimmed. The _pullELFees path in setConsensusLayerData is
  never triggered, so the upper bound check (availableAmountToUpperBound) is only tested via CL rewards.
  - I14 is missing from the numbering — comments reference it but there's no invariant_I14_* function. Either it was deleted or never
  written.```

_Originally posted by @mischat in https://github.com/liquid-collective/liquid-collective-protocol/pull/395#discussion_r3133965118_

### PR DESCRIPTION
**Pull request by iamsahu**
## Summary
Adds four missing invariants and an inline deposit-time check:

- **I7**: Central accounting identity — `InFlightDeposit + totalDepositedActivatedETH == TotalDepositedETH`. This is the most important invariant in the design: every ETH sent to the deposit contract is either still in-flight or has activated.
- **I8**: Per-operator `requestedExits >= exited`. This ordering is load-bearing — `requestETHExits` computes `opPendingExits = requestedExits - exitedETH` and relies on it being non-negative.
- **I9**: `TotalETHExitsRequested >= totalExited`. Every processed exit was either explicitly requested or bumped via the unsolicited path in `_setExitedETH`.
- **I10**: `totalDepositedActivatedETH` monotonically non-decreasing across reports.
- **Deposit share price stability**: Inline check in `sim_deposit` that `totalSupply` and `totalUnderlyingSupply` are unchanged after `depositToConsensusLayerWithDepositRoot` (the whole point of InFlightDeposit).

Closes #417

## Test plan
- [x] `forge test --match-path "contracts/test/accounting/*"` — 39 tests pass
- [x] Fuzz suite exercises all invariants across randomized deposit/exit/slash sequences

## Summary by CodeRabbit

* **Tests**
  * Expanded accounting validation checks to verify accuracy across deposit, activation, and exit operations.
  * Added stricter ordering rules ensuring per-operator and global ETH exit request consistency and activated ETH monotonicity.

### CODE REVIEW

File: contracts/test/accounting/AccountingInvariants.sol [hunk 1]
```diff
@@ -204,4 +210,47 @@ abstract contract AccountingInvariants is BeaconChainSimulator {
         }
         assertEq(totalExited, sum, "I6: exitedETHPerOperator aggregate mismatch");
     }
+
+    /// @notice I7: The central accounting identity — every ETH sent to the deposit contract is
+    ///         either still in-flight or has activated:
+    ///         InFlightDeposit + totalDepositedActivatedETH == TotalDepositedETH
+    function _assertI7_AccountingIdentity() internal {
```
juliaaschmidt: I would give this a more descriptive name, this is only a deposit-related Accounting identity. It does not ocnsider exits or consolidations. I would show that in the invariant name

contracts/test/accounting/AccountingInvariants.sol
```diff
@@ -204,4 +210,47 @@ abstract contract AccountingInvariants is BeaconChainSimulator {
...
+        uint256 inFlight = river.getInFlightDeposit();
+        uint256 activated = river.getLastConsensusLayerReport().totalDepositedActivatedETH;
+        uint256 totalDeposited = river.getTotalDepositedETH();
+        assertEq(
+            inFlight + activated, totalDeposited, "I7: InFlightDeposit + activatedETH != TotalDepositedETH"
+        );
+    }
+
+    /// @notice I8: Per-operator requestedExits >= exited. This ordering is load-bearing —
+    ///         requestETHExits computes opPendingExits = requestedExits - exitedETH and relies
```
juliaaschmidt: this suggests one is in number of exits (requestedExits) and the other is in ETH (exitedETH)?

contracts/test/accounting/AccountingInvariants.sol
```diff
@@ -204,4 +210,47 @@ abstract contract AccountingInvariants is BeaconChainSimulator {
...
+                op.requestedExits,
+                exited,
+                string(abi.encodePacked("I8: op", vm.toString(i), " requestedExits < exited"))
+            );
+        }
+    }
+
+    /// @notice I9: TotalETHExitsRequested >= totalExited. Every processed exit was either
+    ///         explicitly requested or bumped via the unsolicited path in _setExitedETH.
+    function _assertI9_TotalRequestedGeExited() internal {
```
juliaaschmidt, line 263: in your simulated world this might be true. What about slashing though? Would they just be treated as unsolicited exits here? Would this still hold true `Every processed exit was either
    ///         explicitly requested or bumped via the unsolicited path in _setExitedETH.`

mischat:
@mischat-galaxy @julia-galaxy I tried to update this branch against its base branch, but GitHub reported merge conflicts. I have not attempted conflict resolution.

mischat:
@coderabbitai review

mischat:
@coderabbitai review

contracts/test/accounting/AccountingInvariants.sol
```diff
@@ -204,4 +210,47 @@ abstract contract AccountingInvariants is BeaconChainSimulator {
...
+        uint256 opCount = operatorsRegistry.getOperatorCount();
+        uint256[] memory exitedPerOp = operatorsRegistry.getExitedETHPerOperator();
+        for (uint256 i = 0; i < opCount; i++) {
+            OperatorsV3.Operator memory op = operatorsRegistry.getOperator(i);
+            uint256 exited = exitedPerOp.length > i ? exitedPerOp[i] : 0;
+            assertGe(
+                op.requestedExits,
+                exited,
+                string(abi.encodePacked("I8: op", vm.toString(i), " requestedExits < exited"))
+            );
```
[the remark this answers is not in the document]
juliaaschmidt: @copilot this is only meant to assert that the requested exists is great or equal to the exited amount. The assert should not request any more exits here

contracts/test/accounting/AccountingInvariants.sol
juliaaschmidt: addressed in comments

mischat (approved):
OK

PR #507, opened 2026-05-21
feat: consolidation attestation in AttestationVerifierV1

iamsahu:
## Description

Extends `AttestationVerifierV1` with a parallel attestation flow for EIP-7251 consolidation requests. The deposit attestation flow is unchanged at runtime; the deposit entry point is renamed `validate` → `validateDeposits` so the two flows have parallel names.

The consolidation flow has no on-chain buffer. Callers pass a `ConsolidationObject` directly into `validateConsolidation` in calldata. The verifier validates structural shape, builds an EIP-712 digest from the request fields, recovers committee signatures against that digest, and stores the EIP-712 structHash for replay protection.

Integration with `mintLsETHForConsolidation` (currently on `feat/pectra/consolidation-minting`) is intentionally out of scope. This PR is the trust layer only.

### ConsolidationObject

Passed directly into `validateConsolidation`:

1. `address user` — initiator (and eventual recipient of LsETH in the follow-up PR)
2. `bytes[] sourcePubkeys` — 48 bytes each
3. `bytes[] targetPubkeys` — 48 bytes each, paired by index with sources
4. `uint256 totalAmount` — total ETH being consolidated, in wei
5. `bytes[] signatures` — committee EIP-712 ECDSA signatures, 65 bytes each

The first four fields are the EIP-712 typed-data inputs the committee signs. `signatures` is NOT part of the typed data.

### EIP-712

Typehash:
```
AttestConsolidation(address user, bytes[] sourcePubkeys, bytes[] targetPubkeys, uint256 totalAmount)
```

`bytes[]` arrays follow EIP-712 array rules: each element becomes `keccak256(element)`, and the resulting `bytes32[]` is concatenated and hashed (a small `_hashBytesArray` helper).

Cross-flow replay protection — defense-in-depth at two levels:
1. Distinct EIP-712 domain separator (`NAME_HASH = keccak256(\"ConsolidationValidation\")` vs the deposit-side `keccak256(\"DepositToConsensusLayerValidation\")`). Both anchored to River.
2. Distinct typehash from the deposit flow's `Attest(bytes32 depositDataBufferId, bytes32 depositRootHash)`.

### Replay protection

`validateConsolidation` is **state-mutating** (not `view`). On a successful call it records the EIP-712 structHash in `ProcessedConsolidations` and emits `ConsolidationProcessed(structHash)`. A second call with a struct that hashes to the same value reverts with `ConsolidationAlreadyProcessed(structHash)`. The replay check runs **before** the quorum work so a previously-accepted request fails fast even if signatures are still valid.

**DoS caveat** (documented in NatSpec): the function is permissionless, so a malicious caller can front-run a legitimate consumer and mark a request as processed, blocking the legitimate flow. Caller restriction is intentionally deferred — it lives in the eventual River integration (which can gate the verifier or atomically combine validation with its own downstream mint).

### What's added

- `contracts/src/state/attestationVerifier/ConsolidationCommitteeAttesters.sol` — attester mapping + count.
- `contracts/src/state/attestationVerifier/ConsolidationCommitteeAttestationQuorum.sol` — `uint256` quorum.
- `contracts/src/state/attestationVerifier/ConsolidationDomainSeparator.sol` — EIP-712 domain separator anchored to River.
- `contracts/src/state/attestationVerifier/ProcessedConsolidations.sol` — replay-protection mapping `structHash → bool`.
- `contracts/test/components/ConsolidationAttestation.t.sol` — 44 unit tests + `MockRiverAdmin`.

### What's extended

- `contracts/src/AttestationVerifier.1.sol`:
  - **Contract notice** rewritten to describe both flows (deposit + consolidation).
  - `initAttestationVerifierV1` takes 7 params — deposit-side (river, deposit buffer, committee, quorum, fork version) + consolidation-side (committee, quorum) — and configures both flows in one call.
  - `validateDeposits(...)` (was `validate`) — view, deposit attestation + BLS verification over a batch fetched from `DepositDataBuffer`. Logic unchanged; rename only.
  - `validateConsolidation(ConsolidationObject calldata) returns (bool)` — state-mutating. Structural validation → EIP-712 structHash → replay check → ECDSA quorum verification → mark processed + emit. Reverts on any failure; the bool is always `true` on success.
  - `_verifyConsolidationAttestationQuorum(bytes32 digest, bytes[] calldata signatures)` — pure signature-recovery loop using the consolidation committee.
  - `_hashBytesArray` — small EIP-712 `bytes[]` array-hashing helper.
  - Two consolidation-side admin setters (`setConsolidationCommitteeAttester`, `setConsolidationCommitteeAttestationQuorum`) gated by `onlyRiverAdmin`.
  - Four consolidation-side views (`isConsolidationCommitteeAttester`, count, quorum, domain separator).
- `contracts/src/interfaces/IAttestationVerifier.1.sol`:
  - Interface notice rewritten to describe both flows.
  - `ConsolidationObject` struct now lives here.
  - Parallel events, errors, and function signatures for the consolidation flow.
  - `validate` → `validateDeposits` rename mirrored.
- `contracts/src/components/ConsensusLayerDepositManager.1.sol` — the only in-repo caller of `validate` switched to `validateDeposits`.

### Verification

- `forge build --sizes`: `AttestationVerifierV1` runtime size = **12,140 bytes** (12,436-byte EIP-170 margin).
- 44/44 new consolidation tests pass: `forge test --match-contract ConsolidationAttestationTest`.
- Deposit regression: 14/14 deposit attestation tests still pass.
- Full suite: 680/681 pass — the 1 failure is `TlcMigrationTest` failing at `setUp()` for missing `MAINNET_FORK_RPC_URL`, unrelated to this change.

## Notice

- [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?
- [x] Have you added sufficient documentation in your code?

## Pull Request Type

- [x] 💫 New Feature (Non-breaking Change)

## Breaking changes (if applicable)

ABI changes — both safe because this verifier has not been deployed yet:

1. `initAttestationVerifierV1` signature is widened with three consolidation-side params (`_consolidationCommitteeAttesters[]`, `_consolidationQuorum`). The 4th parameter is also renamed from `_quorum` to `_depositQuorum` for symmetry; types are unchanged on the first five slots. All in-repo callers (test harnesses) are updated.
2. The deposit-side entry point is renamed `validate` → `validateDeposits`. The only in-repo caller — `ConsensusLayerDepositManager.depositToConsensusLayerWithAttestation` — is updated in this PR. The function body is unchanged.

The deposit attestation runtime path (committee state, EIP-712 `Attest` typehash, deposit domain separator) is otherwise unchanged.

## Testing

- [x] Have you tested this code with the official test suite?

## Manual tests (if applicable)

N/A — this PR is the trust layer only. End-to-end integration (`mintLsETHForConsolidation`) is the subject of a follow-up PR.

## Release Notes

* **New Features**
  * Added consolidation attestation validation with committee-based verification and replay protection
  * Introduced consolidation committee management and configurable quorum controls
  * Updated consolidation minting to require attestation verification before processing
  * Added external recipient mapping for enhanced consolidation routing

* **Tests**
  * Expanded test coverage for new consolidation verification flows and administrative operations


contracts/src/interfaces/IAttestationVerifier.1.sol
```diff
@@ -6,10 +6,42 @@ import "../libraries/BLS12_381.sol";
...
+    /// @dev    `sourcePubkeys[i]` is consolidated INTO `targetPubkeys[i]` — same-index pairing.
+    ///         `signatures` are the consolidation-committee attestor EIP-712 ECDSA signatures
+    ///         over the typed-data struct
+    ///             AttestConsolidation(address user, bytes[] sourcePubkeys, bytes[] targetPubkeys, uint256 totalAmount)
+    ///         The `signatures` field itself is NOT part of the typed data — only the four
+    ///         request fields are. This is what lets attestors produce signatures over the
+    ///         request without a circular dependency.
+    struct ConsolidationObject {
+        /// @dev Initiator of the consolidation request; eventual recipient of LsETH.
+        address user;
```
juliaaschmidt: Why not withdrawalCredentials? Is that because if custom recipient is set `user != withdrawalCredentials` is possible?

contracts/src/River.1.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity 0.8.34;
 
 import "./interfaces/IAllowlist.1.sol";
 import "./interfaces/IAttestationVerifier.1.sol";
+import "./interfaces/IAttestationVerifier.1.sol";
```
juliaaschmidt: accidental duplicate import here

contracts/src/AttestationVerifier.1.sol
```diff
@@ -83,18 +121,34 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
         address _river,
         address _depositDataBuffer,
         address[] calldata _depositCommitteeAttesters,
-        uint256 _quorum,
-        bytes4 _genesisForkVersion
+        uint256 _depositQuorum,
```
juliaaschmidt: I see what you did here, but I wonder whether there's a way to reduce code duplication?

 The deposit-side and consolidation-side primitives are structurally identical: same storage shape, same operations, same verification loop with the only real difference being the slot string and the EIP-712 typehash.

- ConsolidationCommitteeAttestationQuorum vs DepositCommitteeAttestationQuorum
- ConsolidationCommitteeAttesters vs DepositCommitteeAttesters
- ConsolidationDomainSeparator vs DomainSeparator
- _verifyConsolidationAttestationQuorum vs the existing deposit quorum loop

Would it be worth collapsing each pair into one parametrised  library/function and having the caller pass in which flow it's operating on?

contracts/src/River.1.sol
iamsahu: Fixed.

contracts/src/interfaces/IAttestationVerifier.1.sol
iamsahu: That's true, however as you pointed that we have to work on a WC->address mapping this will have to change to withdrawalCredentials.

contracts/src/AttestationVerifier.1.sol
iamsahu: I will be addressing this in another PR

mischat: I agree, I bet we can clean this up and have less code, if you started or done the work already, that is great. But if not this is something we can do as a follow-up with a tracked issue.

contracts/src/AttestationVerifier.1.sol
```diff
@@ -329,6 +469,79 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+                _hashBytesArray(consolidation.sourcePubkeys),
+                _hashBytesArray(consolidation.targetPubkeys),
+                consolidation.totalAmount
+            )
+        );
+
+        // 4. Replay protection — reject any consolidation we've already accepted.
+        //    Checked BEFORE quorum verification so a previously-accepted request fails fast
+        //    even if the supplied signatures happen to be valid again.
+        if (ProcessedConsolidations.isProcessed(structHash)) {
```
mischat, line 530: **M-1 (medium, non-blocking) — no on-chain source-pubkey uniqueness across requests.**
Replay protection is keyed on the EIP-712 `structHash` of the whole request `(withdrawalAddress, sourcePubkeys[], targetPubkeys[], totalAmount)`. There is no on-chain registry of already-consolidated source pubkeys (the deposit flow has `PectraValidatorPubkeyLookup`). So two *distinct* requests sharing a source — e.g. `{src:[A], amt:32}` and `{src:[A,B], amt:64}` — hash differently, both pass this check, and source `A` is effectively minted twice. The only guard today is the committee not signing overlapping requests. Suggest mirroring `PectraValidatorPubkeyLookup` for consolidated sources, or documenting cross-request source uniqueness as an explicit committee invariant + a test. Fine to address in a follow-up.

contracts/src/AttestationVerifier.1.sol
```diff
@@ -329,6 +469,79 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+        }
+
+        // 3. Compute the EIP-712 digest the committee signed.
+        //    The struct's `signatures` field is NOT part of the typed data — only the four
+        //    request fields are. `bytes[]` arrays follow EIP-712 array rules: each element
+        //    becomes `keccak256(element)`, then the array hashes to `keccak256` over the
+        //    concatenation of those 32-byte element hashes.
+        bytes32 domainSep = ConsolidationDomainSeparator.get();
+        if (domainSep == bytes32(0)) revert ZeroConsolidationDomainSeparator();
+        bytes32 structHash = keccak256(
```
mischat, line 517: **M-2 (low/medium, non-blocking) — replay identity omits `exitEpoch`/nonce.**
The design doc's `consolidationHash` included `exitEpoch`; this struct hashes only `(withdrawalAddress, sourcePubkeys, targetPubkeys, totalAmount)`, so two legitimately-distinct requests sharing those four fields collide (the second is permanently rejected with `ConsolidationAlreadyProcessed`). This lines up with the direction we are already leaning toward — **adding a nonce to the consolidation object** — which would also give each request a unique identity here. Suggest we land that nonce and use it consistently as the replay key across the consolidation (and deposit) flows.

contracts/src/AttestationVerifier.1.sol
```diff
@@ -329,6 +469,79 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+    ///          of (source, target) pairs.
+    ///        - Whether the source validators actually exist on the consensus layer or
+    ///          carry the protocol's withdrawal credentials.
+    ///      These are the responsibility of the caller (off-chain pipeline) and the
+    ///      consolidation committee that signs the request. The eventual
+    ///      `mintLsETHForConsolidation` River integration is the place to enforce any
+    ///      additional financial caps on `totalAmount`.
+    function validateConsolidation(IAttestationVerifierV1.ConsolidationObject calldata consolidation)
+        external
+        onlyRiver
```
mischat, line 489: **L-1 (nit, non-blocking) — PR description's DoS caveat is now stale.**
The description notes this function is permissionless / front-runnable, but it is gated `onlyRiver` here and River gates the caller `onlyKeeper`, so that front-running DoS no longer applies. Worth updating the PR description / NatSpec to match.

contracts/src/River.1.sol
```diff
@@ -180,18 +186,35 @@ contract RiverV1 is
...
+            if (allowlist.isDenied(recipient)) {
+                revert Denied(recipient);
+            }
         }
-        LibSanitize._notZeroAddress(_recipient);
+
+        IAttestationVerifierV1 verifier = IAttestationVerifierV1(AttestationVerifierAddress.get());
+        // Since the verifier validates the consolidation object, we do not validate it here
+        // this reverts if the consolidation is invalid
+        verifier.validateConsolidation(consolidation);
```
mischat, line 212: **L-2 (nit, non-blocking) — return value ignored.**
`validateConsolidation` returns `bool`, but it is ignored here. Harmless (it reverts on failure and always returns `true` on success), so the return is effectively dead — either assert it at the call site or drop the return type for clarity.

contracts/src/River.1.sol
```diff
@@ -180,18 +186,35 @@ contract RiverV1 is
...
+    function mintLsETHForConsolidation(IAttestationVerifierV1.ConsolidationObject calldata consolidation)
+        external
+        onlyKeeper
+    {
+        // we check the allowlist first to fail fast if the withdrawalAddress/recipient is denied
+        IAllowlistV1 allowlist = IAllowlistV1(AllowlistAddress.get());
+        allowlist.onlyAllowed(consolidation.withdrawalAddress, LibAllowlistMasks.CONSOLIDATE_MASK);
+
+        address recipient = IExternalConsolidationRecipientMappingV1(ExternalConsolidationRecipientMappingAddress.get())
+            .getRecipient(consolidation.withdrawalAddress);
```
mischat, line 198: **Cross-cutting (non-blocking) — standardize on the on-chain recipient mapping.**
Resolving the recipient from the on-chain `ExternalConsolidationRecipientMapping` (keyed on the committee-attested `withdrawalAddress`) is the right call and matches the direction we discussed — **standardizing on the on-chain mapping** rather than a backend/off-chain lookup. Flagging so we keep this consistent wherever recipient / withdrawal-credential resolution happens across the consolidation code (naming + lookup pattern), and so the mapping stays the single source of truth.

mischat (commented):
A few non-blocking notes on the consolidation attestation flow — all fine to address in follow-ups. Build + tests verified locally (ConsolidationAttestation 45/45, deposit + River regression 168/168).

contracts/src/AttestationVerifier.1.sol
iamsahu: Created an issue.

contracts/src/AttestationVerifier.1.sol
iamsahu: Created an issue

contracts/src/River.1.sol
iamsahu: Created an issue

contracts/src/River.1.sol
iamsahu: Created an issue

mischat (approved):
LGTM

iamsahu:
@coderabbit review

juliaaschmidt (approved):

mischat:
Tracking the test-coverage gaps from the notes above (consolidation vs deposit attestation parity — cross-request source uniqueness + end-to-end replay/no-double-mint) in #558. Non-blocking for this PR.

PR #540, opened 2026-06-03
Post-pectra validator migration

iamsahu:
## Description

This pull request introduces significant enhancements to support migration and validation of pre-Pectra validator pubkeys, improves the handling of consolidation and withdrawal operations, and refactors imports for consistency. The main changes add new migration functionality, extend interfaces, and strengthen validation logic to ensure both post- and pre-Pectra pubkeys are properly supported throughout the protocol.

**Pre-Pectra Validator Migration and Lookup:**

* Added support for migrating pre-Pectra funded validator pubkeys into the attestation verifier, including a new `migratePrePectraValidatorPubkeys` function, related events, and error handling in `AttestationVerifierV1` and its interface. This allows admins to migrate legacy validator keys in chunks and track their funding status. [[1]](diffhunk://#diff-2cbbaf82cf40abbadf2169784f860b0a4a2fc2bf356e6e03305e081c2a69b90aL452-R489) [[2]](diffhunk://#diff-2cca6fbd0887dba17a65589a8f86c7d3a2a32849fd684cfcbe38c2103be3d779R81-R86) [[3]](diffhunk://#diff-2cca6fbd0887dba17a65589a8f86c7d3a2a32849fd684cfcbe38c2103be3d779R225-R245) [[4]](diffhunk://#diff-2cca6fbd0887dba17a65589a8f86c7d3a2a32849fd684cfcbe38c2103be3d779R327-R334) [[5]](diffhunk://#diff-2cca6fbd0887dba17a65589a8f86c7d3a2a32849fd684cfcbe38c2103be3d779R453-R457)
* Extended `IOperatorsRegistryV1` and its implementation to provide access to legacy funded validator counts and pubkeys for migration, with new methods `getPrePectraFundedValidatorCount` and `getPrePectraValidatorPubkeys`. [[1]](diffhunk://#diff-1fb73f6332a46939cddb1ec200b8f4bbdaac8ef610f23f49c9508b1fba26ac90R252-R267) [[2]](diffhunk://#diff-1459c8eab6cd9b658f70260ee5fd5f17d6c8c1008b8412e52afea55db9804dabR165-R179)

**Attestation and Withdrawal Validation:**

* Updated attestation and withdrawal logic to check both post-Pectra and migrated pre-Pectra pubkeys for funding status, including new error types for unfunded source and target pubkeys, and validation logic in `WithdrawV1`. [[1]](diffhunk://#diff-2cbbaf82cf40abbadf2169784f860b0a4a2fc2bf356e6e03305e081c2a69b90aR527-R530) [[2]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R130-R149) [[3]](diffhunk://#diff-984b02abbb24a29ef36d5b26931fab83ad9df15d29714c4a53269f9f8354b39fR65-R85)
* Renamed the main attestation validation function from `validateDeposits` to `validate` for clarity and future extensibility.

**Refactoring and Import Organization:**

* Refactored imports across several contracts to use shared state files and maintain consistent ordering, improving code clarity and maintainability. [[1]](diffhunk://#diff-34420a68f2d8aba02b8bc93ba4306dcb459547b0c3a4815dfa1737a16c14665eL4-R38) [[2]](diffhunk://#diff-865957c3c5c4b8bbd276c2f684d6877d63be543bdebef39218ba8eaa8b343be0R10-R16) [[3]](diffhunk://#diff-7b338638aeefb6f1e264218c49aab994b7c0b9b06328f2d95b74a82b9a9cc702L16-R23) [[4]](diffhunk://#diff-2cbbaf82cf40abbadf2169784f860b0a4a2fc2bf356e6e03305e081c2a69b90aR11-R12) [[5]](diffhunk://#diff-2cbbaf82cf40abbadf2169784f860b0a4a2fc2bf356e6e03305e081c2a69b90aR26) [[6]](diffhunk://#diff-1459c8eab6cd9b658f70260ee5fd5f17d6c8c1008b8412e52afea55db9804dabR16)

These changes collectively ensure robust support for both legacy and new validator sets, streamline migration, and enhance the protocol's safety and maintainability.

## Notice

## Pull Request Type

## Testing


contracts/src/AttestationVerifier.1.sol
```diff
@@ -449,7 +452,41 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+        }
+
+        bytes[] memory pubkeys = operatorsRegistry.getPrePectraValidatorPubkeys(operatorIndex, startIndex, stopIndex);
+        uint256 len = pubkeys.length;
+        for (uint256 i = 0; i < len; ++i) {
+            bytes memory pubkey = pubkeys[i];
+            if (pubkey.length != DEPOSIT_PUBKEY_LENGTH) {
+                revert InvalidPrePectraMigrationPubkeyLength(operatorIndex, startIndex + i, pubkey.length);
+            }
+            PrePectraValidatorPubkeyLookup.add(pubkey);
```
juliaaschmidt, line 487: if we had to rerun the function for the same operator multiple times, it would simply read them again here right?

contracts/src/OperatorsRegistry.1.sol
```diff
@@ -161,6 +162,21 @@ contract OperatorsRegistryV1 is IOperatorsRegistryV1, Initializable, Administrab
...
+        return OperatorsV2.get(operatorIndex).funded;
+    }
+
+    /// @inheritdoc IOperatorsRegistryV1
+    function getPrePectraValidatorPubkeys(uint256 operatorIndex, uint256 startIndex, uint256 stopIndex)
+        external
+        view
+        returns (bytes[] memory publicKeys)
+    {
+        OperatorsV2.get(operatorIndex);
```
juliaaschmidt: pretty neat!

juliaaschmidt (commented):
Looks good to me overall. 👍  Sent over the AI findings though, please review.

contracts/src/AttestationVerifier.1.sol
iamsahu: That's right

contracts/src/state/attestationVerifier/PrePectraValidatorPubkeyLookup.sol
```diff
@@ -0,0 +1,42 @@
...
+pragma solidity 0.8.34;
+
+import "../../libraries/LibUnstructuredStorage.sol";
+
+/// @title PrePectraValidatorPubkeyLookup
+/// @notice Unstructured-storage mapping of pre-Pectra pubkeys migrated from the legacy
+///         operators-registry ValidatorKeys storage into AttestationVerifier state.
+library PrePectraValidatorPubkeyLookup {
+    bytes32 internal constant PRE_PECTRA_VALIDATOR_PUBKEY_LOOKUP_MAPPING_BASE_SLOT =
+        bytes32(uint256(keccak256("attestationVerifier.state.prePectraValidatorPubkeyLookup.mapping")) - 1);
```
mischat: rename these later, make consistent ...

contracts/src/AttestationVerifier.1.sol
```diff
@@ -459,6 +462,60 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+        }
+
+        emit MigratedPrePectraValidatorPubkeys(operatorIndex, startIndex, stopIndex);
+    }
+
+    /// @inheritdoc IAttestationVerifierV1
+    function removePrePectraValidatorPubkey(bytes[] calldata pubkeys) external onlyRiverAdmin {
```
mischat: this takes an array ... please rename from singular  ... one for later ...

contracts/src/interfaces/IAttestationVerifier.1.sol
```diff
@@ -84,28 +85,36 @@ interface IAttestationVerifierV1 {
     /// @param stopIndex The exclusive stop key index
     event MigratedPrePectraValidatorPubkeys(uint256 indexed operatorIndex, uint256 startIndex, uint256 stopIndex);
 
+    /// @notice Emitted when a chunk of pre-Pectra validator pubkeys is removed from verifier state.
```
juliaaschmidt, line 88: from the verifier state? That comment is confusing. This removal has nothing to do with the Verifier flow.

contracts/src/interfaces/IAttestationVerifier.1.sol
```diff
@@ -239,9 +253,25 @@ interface IAttestationVerifierV1 {
     /// @param length The observed pubkey length
     error InvalidPrePectraMigrationPubkeyLength(uint256 operatorIndex, uint256 keyIndex, uint256 length);
 
-    /// @notice A target pubkey is not funded
-    /// @param pubkey The offending 48-byte BLS pubkey
-    error TargetPubkeyNotFunded(bytes pubkey);
+    /// @notice The pre-Pectra removal batch is empty.
+    error InvalidPrePectraRemovalEmptyPubkeys();
```
juliaaschmidt, line 263: Ideally, I would move all the Migration-related functionality, errors and events to a separate migration interface instead of mixing it with the Attestation verifier logic. Maybe we can create an issue for it?

contracts/src/AttestationVerifier.1.sol
```diff
@@ -459,6 +463,89 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+    function validateSelfConsolidation(bytes[] calldata pubkeys, uint256)
+        external
+        payable
+        onlyRiver
+        returns (IWithdrawV1.ConsolidationRequest[] memory)
+    {
+        uint256 len = pubkeys.length;
+        if (len == 0) {
+            revert InvalidSelfConsolidationEmptyPubkeys();
+        }
```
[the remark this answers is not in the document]
iamsahu: Will fix this

contracts/src/OperatorsRegistry.1.sol
```diff
@@ -163,6 +164,23 @@ contract OperatorsRegistryV1 is IOperatorsRegistryV1, Initializable, Administrab
...
+    function getPrePectraValidatorPubkeys(uint256 operatorIndex, uint256 startIndex, uint256 stopIndex)
+        external
+        view
+        returns (bytes[] memory publicKeys)
+    {
+        OperatorsV2.Operator storage op = OperatorsV2.get(operatorIndex);
+        if (stopIndex > op.funded) revert PrePectraRangeExceedsFunded(operatorIndex, stopIndex);
+        if (startIndex >= stopIndex) revert InvalidPrePectraRange(operatorIndex, startIndex, stopIndex);
+        (publicKeys,) = ValidatorKeys.getKeys(operatorIndex, startIndex, stopIndex - startIndex);
+    }
```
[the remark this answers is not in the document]
iamsahu: Bubbling works here as adding more checks will increase gas costs.

contracts/src/AttestationVerifier.1.sol
```diff
@@ -459,6 +463,89 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+
+        emit MigratedPrePectraValidatorPubkeys(operatorIndex, startIndex, stopIndex);
+    }
+
+    /// @inheritdoc IAttestationVerifierV1
+    function validateSelfConsolidation(bytes[] calldata pubkeys, uint256)
+        external
+        payable
+        onlyRiver
```
mischat: **SC-1 (LOW) — unused `maxFeePerConsolidation` param is dead/misleading surface**

_(Updated: the `payable` modifier was correctly dropped in `00514ec0` — thanks. One leftover remains.)_

`validateSelfConsolidation(bytes[] calldata pubkeys, uint256)` still takes a `maxFeePerConsolidation` argument that is never read. The actual fee bound is enforced downstream in `Withdraw._validateAndReturnFee`, so this parameter does nothing here.

Suggest either removing the param, or actually using it (e.g. to fail fast before building the requests), so the signature doesn't imply a fee responsibility this function doesn't have.

contracts/src/AttestationVerifier.1.sol
```diff
@@ -459,6 +463,89 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
+            bytes calldata pubkey = pubkeys[i];
+            if (pubkey.length != DEPOSIT_PUBKEY_LENGTH) {
+                revert InvalidSelfConsolidationPubkeyLength(i, pubkey.length);
+            }
+            if (!PrePectraValidatorPubkeyLookup.isPubkeyFunded(pubkey)) {
+                revert PrePectraValidatorPubkeyNotFunded(pubkey);
+            }
+            requests[i] = IWithdrawV1.ConsolidationRequest({srcPubkeys: new bytes[](1), targetPubkey: pubkey});
+            requests[i].srcPubkeys[0] = pubkey;
+        }
```
mischat, line 518: **SC-2 (INFO / design clarification) — upgraded legacy validators are never promoted to the post-Pectra lookup**

This loop builds a `src == target` self-consolidation request (the on-chain 0x01→0x02 compounding upgrade) for each migrated pre-Pectra pubkey, but the pubkey is never added to `PectraValidatorPubkeyLookup` — it stays only in the pre-Pectra lookup.

Consequence: after a successful self-consolidation, the validator is 0x02 on the beacon chain but still fails the **normal** target check in `Withdraw.consolidate` (which requires `isPubkeyFunded`, i.e. post-Pectra-lookup membership). So an upgraded legacy validator can only ever be a *source* or be self-consolidated again — never a normal consolidation *target*. Merging multiple validators must therefore target a freshly-funded post-Pectra validator, not an upgraded legacy one.

Is that the intended lifecycle? If the strategy is that validators are only ever consolidated *into* freshly-funded post-Pectra validators, this is fine — just confirming. If upgraded legacy validators should be usable as targets, consider `PectraValidatorPubkeyLookup.add(pubkey)` here on self-consolidation (and clearing the pre-Pectra entry).

contracts/src/Withdraw.1.sol
```diff
@@ -132,31 +132,26 @@ contract WithdrawV1 is IWithdrawV1, Initializable, ReentrancyGuard, IProtocolVer
...
-                }
+            bytes calldata targetPubkey = requests[i].targetPubkey;
+            _validatePubkeyLength(targetPubkey);
+
+            bool isTargetFunded = attestationVerifier.isPubkeyFunded(targetPubkey);
+            bool isSelfConsolidation =
+                requests[i].srcPubkeys.length == 1 && keccak256(requests[i].srcPubkeys[0]) == keccak256(targetPubkey);
+
+            if (
+                !isTargetFunded && (!isSelfConsolidation || !_isKnownValidatorPubkey(attestationVerifier, targetPubkey))
```
juliaaschmidt: why not simply `if !_isKnownValidatorPubkey(attestationVerifier, targetPubkey)` here?

juliaaschmidt (commented):
When do we add the 0x01-> 0x02 migrated pubkeys gets added to the important PectraValidator mapping?

contracts/src/AttestationVerifier.1.sol
```diff
@@ -459,6 +463,80 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1 {
...
```
mischat, line 518: **SC-3 (LOW) — no dedup of `pubkeys`; duplicates waste fee**

This loop builds one self-consolidation request per entry without checking for repeats. If `pubkeys` contains the same pubkey twice, two identical EL self-consolidation requests are emitted downstream and `fee` is charged for each (`Withdraw.consolidate` counts `totalNumOfConsolidationOperations` across all requests). The second is effectively a redundant no-op upgrade.

No protocol fund risk — value is keeper-funded, `selfConsolidation` is `onlyKeeper`, and excess is refunded to the keeper. But it silently wastes fee on the duplicate. Consider an in-batch dedup (revert on repeated pubkey) or document that callers must pass unique pubkeys.

contracts/src/AttestationVerifier.1.sol
mischat: @juliaaschmidt and @iamsahu this one seems like it is worth fixing.

mischat (commented):
I left a few comments, perhaps we should go through the comments and decide what to create issues for, what needs fixing now, and what needs closing...

contracts/src/AttestationVerifier.1.sol
juliaaschmidt: Yes, I agree with this finding.I have already pinged prafful about it too ;)

contracts/src/AttestationVerifier.1.sol
juliaaschmidt: this is true but when I raised no dedupe checks in the past @iamsahu said we can just assume our keeper/consolidator will pass sane input. It is not a public function. I am even thinking of removing the dedupe checks in the deposit/topup flow and trust the committee?

contracts/src/AttestationVerifier.1.sol
juliaaschmidt: Addressed in https://github.com/liquid-collective/liquid-collective-protocol/pull/540/commits/d00754d139216f307c8987f046dd4ff3f328f11c

contracts/src/AttestationVerifier.1.sol
mischat: This is actually fixed by your commit d00754d1 ...

this is how : 

```
The original SC-3 concern

  validateSelfConsolidation builds one consolidation request per entry in pubkeys with no uniqueness check. If a keeper passed the same
  pubkey twice, you'd get two identical EL self-consolidation requests and pay fee twice — wasted fee, no fund risk.

  What d00754d1 added (for SC-2)

  Lines 514–515 now promote each pubkey from the pre-Pectra lookup to the post-Pectra lookup inside the loop:

  if (!PrePectraValidatorPubkeyLookup.isPubkeyFunded(pubkey)) {  // line 511 — gate
      revert PrePectraValidatorPubkeyNotFunded(pubkey);
  }
  PrePectraValidatorPubkeyLookup.remove(pubkey);                 // line 514 — consumes the gate
  PectraValidatorPubkeyLookup.add(pubkey);                       // line 515

  The intent of those two lines was SC-2 (reflect the 0x01→0x02 upgrade). But notice the gate at line 511 reads pre-Pectra membership, and
  line 514 removes that membership in the same iteration.

  Why that incidentally kills duplicates

  The pre-Pectra check at 511 is now effectively a one-shot, consumable gate per pubkey:

  - Iteration A (first occurrence): pubkey is in the pre-Pectra lookup → passes line 511 → line 514 removes it.
  - Iteration B (duplicate, same batch): pubkey is no longer in the pre-Pectra lookup → line 511 is now true → reverts
  PrePectraValidatorPubkeyNotFunded.

  So a batch containing a duplicate now reverts wholesale instead of silently emitting a second request. The same logic also blocks
  re-consolidating an already-upgraded validator across separate transactions.

  Why I called it a "by-product" and not a real fix

  It's a behavioral consequence of the SC-2 promotion, not a dedup that was designed in. That matters because:

  - It only protects the validateSelfConsolidation path. The plain River.consolidate → Withdraw.consolidate path doesn't go through this
  promotion, so duplicate handling there is unchanged (SC-3 wasn't really about that path, but worth noting the protection is local, not
  global).
  - The revert reason is PrePectraValidatorPubkeyNotFunded — semantically "not a known pre-Pectra key," which is misleading for the
  duplicate case. A caller debugging a duplicate batch gets an error that reads like "this key was never migrated," not "you listed it
  twice."
  - It's an implicit invariant. If someone later refactors the promotion (e.g. moves the remove/add after the loop, or makes promotion
  idempotent), the dedup protection silently disappears with no test guarding it.

  So: SC-3's symptom (wasted fee on duplicates) is genuinely gone for the self-consolidation path — but it's an emergent side effect of the
  SC-2 change, with a confusing revert message and no dedicated test, rather than an intentional, well-named dedup guard. If you want it
  to be a real fix, an explicit in-batch uniqueness check (or a clearer revert) plus a test would lock it in.
```

We could create another issue to make the revert more useful ... the act of adding the pubkey to the mapping, essentially, dedupes ...

contracts/src/interfaces/IAttestationVerifier.1.sol
mischat: seems like a good idea to move to an issue.

mischat (approved):
I am happy with this, all of my comments have been addressed.

contracts/src/interfaces/IAttestationVerifier.1.sol
juliaaschmidt: opened #579

contracts/src/interfaces/IAttestationVerifier.1.sol
juliaaschmidt: opened #578

contracts/src/Withdraw.1.sol
mischat: Not equivalent — the `isSelfConsolidation` gate is load-bearing, so we can't drop it.

Let `F = isPubkeyFunded(target)` (post-Pectra / 0x02), `P = isPrePectraValidatorPubkeyFunded(target)` (pre-Pectra / 0x01), `S = isSelfConsolidation`.

- **Current check** passes iff `F || (S && P)`.
- **`!_isKnownValidatorPubkey(target)`** passes iff `F || P`.

They diverge in exactly one case — `P && !S`: a pre-Pectra (0x01) target with **foreign** sources (not a self-consolidation):

| case | current | proposed |
|------|---------|----------|
| `F` (0x02 target) | ✅ | ✅ |
| `S && P` (self-consolidating a 0x01 key → the upgrade) | ✅ | ✅ |
| **`P && !S` (foreign sources → 0x01 target)** | ⛔ revert | ✅ pass |

The EL requires a consolidation **target to be 0x02 (compounding)**. A 0x01 validator only becomes a valid target by self-consolidating (`src == target`), which is the on-chain 0x01→0x02 upgrade — that's the `S && P` allowance. Consolidating *other* validators into a still-0x01 target is invalid; the current check rejects it, the simplified one would let it through.

That said, the reuse of `_isKnownValidatorPubkey` here does read awkwardly: when this line runs, `!isTargetFunded` is already known, so its `isPubkeyFunded` disjunct is always false (a wasted SLOAD) and it collapses to `isPrePectraValidatorPubkeyFunded`. A clearer, semantically-identical form:

```solidity
if (!isTargetFunded && !(isSelfConsolidation && attestationVerifier.isPrePectraValidatorPubkeyFunded(targetPubkey))) {
    revert TargetPubkeyNotFunded(targetPubkey);
}
```

i.e. "target must be post-Pectra funded, or be a self-consolidation of a known pre-Pectra key."

(Note: post-`d00754d1`, the `selfConsolidation` path promotes the pre-Pectra key into the post-Pectra lookup before `Withdraw` runs, so `F` already holds there and `S && P` isn't exercised on that path — it's still reachable via the plain `consolidate` path, which doesn't promote.)

contracts/src/AttestationVerifier.1.sol
juliaaschmidt: We assume that self consolidation risk of failure is less likely than onto other validators

juliaaschmidt (approved):

PR #634, opened 2026-06-23
test(accounting): assert activeCLETH == 0 after V2->V3 migration

CWagamanEure:
## What

Adds three assertions to the V2→V3 migration test confirming `activeCLETH == 0` for every migrated operator.

## Why

The migration resets `activeCLETH` to 0 (`OperatorsRegistry.1.sol`), and that zero is **spec-critical**: until the first post-upgrade oracle report lands, a non-zero `activeCLETH` would let exits be allocated against unconfirmed consensus-layer balance — the source comment flags this as an exit-deadlock risk. The test already verified the funded/exited ×32-ETH scaling but never checked `activeCLETH`, so a migration that left it non-zero would have passed on coverage alone.

## Changes

- `testMigrationV2toV3`: assert `activeCLETH == 0` for both operators.
- `testMigrationSingleOperatorNoStops`: assert `activeCLETH == 0` for the operator.

## Testing

- 4/4 `MigrationTest` pass.
- Full suite: 987 passed; the 1 failure is the pre-existing `TlcMigrationTest` that needs `MAINNET_FORK_RPC_URL` (unrelated).

Surfaced by the accounting test-adequacy audit (finding F-16).

### DIFF
--- a/contracts/test/accounting/scenarios/Migration.t.sol
+++ b/contracts/test/accounting/scenarios/Migration.t.sol
@@ -96,10 +96,14 @@ contract MigrationTest is Test {
         OperatorsV3.Operator memory v3Op0 = registry.getOperator(0);
         assertEq(v3Op0.funded, uint256(op0Funded) * 32 ether, "op0 funded ETH");
         assertEq(v3Op0.active, true, "op0 active");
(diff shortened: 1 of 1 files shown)


mischat (dismissed):
:LGTM

juliaaschmidt (approved):

PR #639, opened 2026-06-30
style: refactor the consolidation target-validation check in `WithdrawV1._processConsolidationRequest` for readability

### LINKED ISSUE #580: style: make condition more readable
Not equivalent — the `isSelfConsolidation` gate is load-bearing, so we can't drop it.

Let `F = isPubkeyFunded(target)` (post-Pectra / 0x02), `P = isPrePectraValidatorPubkeyFunded(target)` (pre-Pectra / 0x01), `S = isSelfConsolidation`.

- **Current check** passes iff `F || (S && P)`.
- **`!_isKnownValidatorPubkey(target)`** passes iff `F || P`.

They diverge in exactly one case — `P && !S`: a pre-Pectra (0x01) target with **foreign** sources (not a self-consolidation):

| case | current | proposed |
|------|---------|----------|
| `F` (0x02 target) | ✅ | ✅ |
| `S && P` (self-consolidating a 0x01 key → the upgrade) | ✅ | ✅ |
| **`P && !S` (foreign sources → 0x01 target)** | ⛔ revert | ✅ pass |

The EL requires a consolidation **target to be 0x02 (compounding)**. A 0x01 validator only becomes a valid target by self-consolidating (`src == target`), which is the on-chain 0x01→0x02 upgrade — that's the `S && P` allowance. Consolidating *other* validators into a still-0x01 target is invalid; the current check rejects it, the simplified one would let it through.

That said, the reuse of `_isKnownValidatorPubkey` here does read awkwardly: when this line runs, `!isTargetFunded` is already known, so its `isPubkeyFunded` disjunct is always false (a wasted SLOAD) and it collapses to `isPrePectraValidatorPubkeyFunded`. A clearer, semantically-identical form:

```solidity
if (!isTargetFunded && !(isSelfConsolidation && attestationVerifier.isPrePectraValidatorPubkeyFunded(targetPubkey))) {
    revert TargetPubkeyNotFunded(targetPubkey);
}
```

i.e. "target must be post-Pectra funded, or be a self-consolidation of a known pre-Pectra key."

(Note: post-`d00754d1`, the `selfConsolidation` path promotes the pre-Pectra key into the post-Pectra lookup before `Withdraw` runs, so `F` already holds there and `S && P` isn't exercised on that path — it's still reachable via the plain `consolidate` path, which doesn't promote.)

_Originally posted by @mischat in https://github.com/liquid-collective/liquid-collective-protocol/pull/540#discussion_r3379654425_

mischat:
## Notice

This pull request updates the withdrawal validation logic to better handle consolidation scenarios for validator pubkeys, especially around the Pectra upgrade. The main changes clarify when a target pubkey is considered valid and introduce a helper function for checking pre-Pectra funding status.

### Withdrawal target validation improvements

* Updated the logic for validating consolidation targets in `WithdrawV1` so that a target pubkey is valid if it is either already funded post-Pectra (0x02), or is a known pre-Pectra (0x01) funded key being self-consolidated (i.e., during the on-chain 0x01 -> 0x02 upgrade). This ensures correct handling of validator pubkeys during protocol upgrades.

### Helper function addition

* Added the `_isPrePectraFunded` internal helper function to encapsulate the check for whether a pubkey is a known pre-Pectra (0x01) funded validator, improving code clarity and reuse.

## Pull Request Type

## Testing

### DIFF
--- a/contracts/src/Withdraw.1.sol
+++ b/contracts/src/Withdraw.1.sol
@@ -155,8 +155,8 @@ contract WithdrawV1 is IWithdrawV1, Initializable, ReentrancyGuard, IProtocolVer
 
         // A consolidation target must be post-Pectra funded (0x02), or be the self-consolidation
         // of a known pre-Pectra (0x01) key — i.e. the on-chain 0x01 -> 0x02 upgrade.
-        bool isValidTarget =
-            isTargetFunded || (isSelfConsolidation && _isPrePectraFunded(attestationVerifier, targetPubkey));
+        bool isValidTarget = isTargetFunded
+            || (isSelfConsolidation && _isKnownPrePectraValidatorPubkey(attestationVerifier, targetPubkey));
 
         if (!isValidTarget) {
             revert TargetPubkeyNotFunded(targetPubkey);
@@ -225,13 +225,13 @@ contract WithdrawV1 is IWithdrawV1, Initializable, ReentrancyGuard, IProtocolVer
     }
 
     /// @notice Internal: check whether a pubkey is a known pre-Pectra (0x01) funded validator
-    function _isPrePectraFunded(IAttestationVerifierV1 attestationVerifier, bytes calldata pubkey)
+    function _isKnownPrePectraValidatorPubkey(IAttestationVerifierV1 attestationVerifier, bytes calldata pubkey)
         internal
         view
         returns (bool)
     {
-        return IAttestationVerifierPectraMigrationV1(address(attestationVerifier))
-            .isPrePectraValidatorPubkeyFunded(pubkey);
+        return
+            IAttestationVerifierPectraMigrationV1(address(attestationVerifier)).isPrePectraValidatorPubkeyFunded(pubkey);
     }
 
     /// @notice Internal: refund excess fee to recipient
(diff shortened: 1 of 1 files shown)


contracts/src/Withdraw.1.sol
```diff
@@ -219,6 +224,16 @@ contract WithdrawV1 is IWithdrawV1, Initializable, ReentrancyGuard, IProtocolVer
         return attestationVerifier.isPubkeyFunded(pubkey) || pectraMigration.isPrePectraValidatorPubkeyFunded(pubkey);
     }
 
+    /// @notice Internal: check whether a pubkey is a known pre-Pectra (0x01) funded validator
+    function _isPrePectraFunded(IAttestationVerifierV1 attestationVerifier, bytes calldata pubkey)
```
juliaaschmidt: Tiny naming nit (non-blocking): could we rename `_isPrePectraFunded` → `_isKnownPrePectraValidatorPubkey` to mirror `_isKnownValidatorPubkey`? They're effectively a set and its subset, so matching the `_isKnown…ValidatorPubkey` shape makes that relationship read clearly at the call site. The current name diverges a bit on both the
  known/funded wording and the dropped ValidatorPubkey suffix. Happy either way, just a consistency thing.

juliaaschmidt (dismissed):

juliaaschmidt (approved):
LGTM 🥇

PR #645, opened 2026-07-02
feat: mark DepositDataBuffer batch processed from River after attested deposit (BS-4394)

### LINKED ISSUE #621: feat: add river update call to `DepositDataBuffer` to mark `DepositDataBufferObject` as processed
When we perform the deposit of the DepositdataObject we read from the deposit data buffer, we need to mark the deposit data batch as processed = True so we can never handle it again accidentally. we do so by a cross-contract call to the DepositDataBuffer contract from River every time we successfully perform depositToConsensusLayerWithAttestation of a batch.

We already have a mapping for the depositDataBufferId to store so we can never accidentally handle the same one twice too.

juliaaschmidt:
## Summary

Closes #621.

After River successfully executes `depositToConsensusLayerWithAttestation` for a batch it read from the on-chain `DepositDataBuffer`, River now makes a **cross-contract call to the buffer** to flip that batch's `processed` flag — so the buffer itself can never re-serve/re-queue the same batch. This adopts the buffer-side interface added in [AlluvialFinance/frontrun-mitigation#32](https://github.com/AlluvialFinance/frontrun-mitigation/pull/32) (`markDepositDataProcessed(bytes32)`, River-gated, emits `DepositDataProcessed`).

The existing **local** replay guard is unchanged and remains the primary in-tx protection: `AttestationVerifier` still marks processed ids in `ProcessedDepositDataBufferIds` via `verifier.markDepositDataBufferIdProcessed(...)`. This PR adds the missing *outward* notification to the buffer.

## Changes

- **`IDepositDataBuffer.sol`** — add `markDepositDataProcessed(bytes32)`, `isDepositDataProcessed(bytes32)`, and the `DepositDataProcessed` event.
- **`ConsensusLayerDepositManager.1.sol`** — new step **5b** in `depositToConsensusLayerWithAttestation`: `IDepositDataBuffer(verifier.getDepositDataBuffer()).markDepositDataProcessed(depositDataBufferId)`. The call originates from the River contract (so `msg.sender` at the buffer is the River proxy, matching the buffer's `RIVER` gate) and the buffer address is sourced from the verifier — no new River storage/init/setter. A buffer-side revert (unknown / already-processed) aborts the whole deposit (**hard revert**, no swallowing).
- **Test mocks** — implement the new interface members across the three `IDepositDataBuffer` mocks (`River.1.t.sol`, `ConsensusLayerDepositManagerAttestation.t.sol`, `AccountingHarnessBase.sol`).
- **Tests** — `testSuccessfulDeposit_marksBufferProcessed` (flag flips + `DepositDataProcessed` emitted) and `testRevert_bufferMarkRejectionRevertsDeposit` (buffer revert propagates, no deposits execute).

## Notes / scope

- **Interface divergence (out of scope):** the LC repo's `IDepositDataBuffer.sol` uses a different data model and `depositDataBufferId` derivation than the frontrun-mitigation buffer. This PR only adopts the `markDepositDataProcessed(bytes32)` method signature and stays self-consistent within this repo; reconciling the two data models is a separate integration concern.
- **Deployment:** the deployed buffer's immutable `RIVER` must be set to the River proxy address (a frontrun-mitigation deployment concern).

## Testing

- `forge test --match-path "contracts/test/components/ConsensusLayerDepositManagerAttestation.t.sol"` → 106 passed
- `forge test --match-contract RiverV1` → 116 passed
- Accounting suites → 57 passed
- `forge fmt --check` clean

### DIFF
--- a/contracts/src/components/ConsensusLayerDepositManager.1.sol
+++ b/contracts/src/components/ConsensusLayerDepositManager.1.sol
@@ -163,8 +163,16 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
             depositDataBufferId, depositRootHash, signatures, depositContract, withdrawalCredentials, committedBalance
         );
 
-        // 5. Mark the batch ID processed BEFORE any external interactions.
+        // 5. Mark the batch ID processed in River's local replay guard before any external
+        //    interactions. This is the primary in-tx guard against double-processing.
         verifier.markDepositDataBufferIdProcessed(depositDataBufferId);
+
+        // 5b. Notify the external DepositDataBuffer to flip its own `processed` flag so the
+        //     batch can never be re-served/re-queued from the buffer side. Gated to River on
+        //     the buffer (msg.sender == this River contract). The buffer address is sourced
+        //     from the AttestationVerifier, the single owner of that dependency. A revert here
+        //     (unknown / already-processed batch) reverts the whole deposit.
+        IDepositDataBuffer(verifier.getDepositDataBuffer()).markDepositDataProcessed(depositDataBufferId);
 
         // 6. Update operator funded validator accounting
         _updateFundedETHFromBuffer(batch.deposits, batch.topUps);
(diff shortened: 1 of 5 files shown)


juliaaschmidt:
This work is now already included in PR #652

PR #651, opened 2026-07-06
test: cover consolidation reward surplus reporting

### LINKED ISSUE #574: Test gap: BeaconChainSimulator can't produce reported > buffered consolidation divergence (clamp reward-surplus path untested end-to-end)
## Summary
PR #516 fixes the HIGH-liveness consolidation-reporting bug by replacing the hard revert with a clamp (`bufferReduction = min(increase, buffer)`, surplus → APR-bounded rewards) in `OracleManager.1.sol` (commit 75c5c41a, lines ~350-356). The clamp's reward-surplus path is now **unit-fuzzed**, but it is **not exercised end-to-end through the realistic beacon-chain simulator**, because the simulator cannot produce a report where `totalExternalConsolidationsAmountReported > ConsolidationBuffer`.

## The gap
`BeaconChainSimulator._buildReport` derives both legs from the same `_simConsolidatedBalance`:

```solidity
// contracts/test/accounting/BeaconChainSimulator.sol:258
report.validatorsBalance = validatorsBalance + _simConsolidatedBalance;
// contracts/test/accounting/BeaconChainSimulator.sol:263
report.totalExternalConsolidationsAmountReported = _simConsolidatedBalance;
```

So every simulator-built report has `reported == buffered`, and the `reported > buffered` branch (the one the clamp exists to handle) can never arise naturally. The scenario tests in `contracts/test/accounting/scenarios/ConsolidationCoverage.t.sol` (`testConsolidationReportReducesBufferAndDoesNotIncreaseAssets`, `testConsolidationReportWithValidatorBalanceIncreaseNetsOut`) only cover `delta <= buffer` (1 ≤ 4 and 1 == 1), and obtain even that via manual `_buildBadReport` struct mutation.

## Why it matters
The exact production scenario the clamp protects against is: a **source validator keeps accruing CL rewards while sitting in the exit queue**, so `process_pending_consolidations` moves `min(actual_balance, effective_balance)` — which can land **~1 ETH+ above** the committee-attested amount that was minted/buffered at request time (an EB hysteresis boundary). That divergence is what used to brick all oracle reporting. We have unit coverage of the arithmetic, but no realistic end-to-end regression test that this divergence books the surplus as APR-bounded rewards and does **not** revert.

## Recommendation
- Add a simulator knob that decouples the two legs — e.g. `_simConsolidationRewardSurplus` that bumps `report.validatorsBalance` **without** bumping `report.totalExternalConsolidationsAmountReported` (model the source validator earning in the exit queue post-request).
- Add a scenario test at an EB hysteresis boundary asserting: buffer drawn to 0, the surplus `reported_actual − buffered` surfaces in `validatorsBalance` as rewards, gated by `TotalValidatorBalanceIncreaseOutOfBound`, and reporting does **not** revert.
- Optionally a paired test where the surplus exceeds the APR ceiling and the report reverts with `TotalValidatorBalanceIncreaseOutOfBound` (so the boundary is pinned from both sides).

## Acceptance criteria
- [ ] The simulator can build a consistent report with `totalExternalConsolidationsAmountReported < validatorsBalance`-derived consolidation principal (reported < actual landed).
- [ ] An end-to-end scenario exercises `reported > buffered` through the real River → Oracle flow (not just a unit `setConsensusLayerData` call) and asserts the no-revert reward-netting behaviour.
- [ ] A boundary test pins the APR-ceiling revert case.

## Notes
Found during differential review of #516. The code fix (clamp) and the test-intent fix (the brick test was converted to a clamp/netting fuzz test) are resolved — see review threads on `OracleManager.1.sol:349` and `ConsolidationCoverage.t.sol:142`. This issue tracks only the remaining **simulator/scenario** coverage gap raised on the `BeaconChainSimulator.sol:263` thread: https://github.com/liquid-collective/liquid-collective-protocol/pull/516#discussion_r3365813670

iamsahu:
## Summary
- Add a simulator path for external consolidations where the buffered principal, reported landed amount, and validator-balance-only surplus can diverge.
- Include the surplus in simulator reward accounting while keeping the reported consolidation total distinct.
- Add end-to-end ConsolidationCoverage scenarios for the no-revert reward-netting path and the APR-ceiling revert boundary.

## Test Plan
- forge fmt --check contracts/test/accounting/BeaconChainSimulator.sol contracts/test/accounting/scenarios/ConsolidationCoverage.t.sol
- git diff --check
- forge test --match-contract ConsolidationCoverageScenarioTest -vvv
- forge test

Closes #574

### DIFF
--- a/contracts/test/accounting/BeaconChainSimulator.sol
+++ b/contracts/test/accounting/BeaconChainSimulator.sol
@@ -55,10 +55,6 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
     ///      Monotonically increasing — incremented by the consolidation report step. `_buildReport` reports it
     ///      as totalExternalConsolidationsAmountReported and adds it to validatorsBalance.
     uint256 internal _simConsolidatedBalance;
-    /// @dev Cumulative CL rewards that landed above the externally reported consolidation amount.
-    ///      Added to validatorsBalance, but not to totalExternalConsolidationsAmountReported, so scenario tests
-    ///      can model source validators earning while queued before consolidation processing.
-    uint256 internal _simConsolidationRewardSurplus;
 
     uint256 internal _lastReportedSkimmed;
     uint256 internal _lastReportedExited;
@@ -235,20 +231,13 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
     ///      `bufferedPrincipal` is the principal minted into the consolidation buffer at request time (counted in
     ///      `_simTotalUserDeposited` for I2 ETH-conservation).
     ///      `reportedAmount` is the delta to add to the cumulative `totalExternalConsolidationsAmountReported` for this report.
-    ///      `consolidationRewardSurplus` is extra CL balance above `reportedAmount` that lands in validatorsBalance (counted as rewards).
-    function sim_reportExternalConsolidation(
-        uint256 bufferedPrincipal,
-        uint256 reportedAmount,
-        uint256 consolidationRewardSurplus
-    ) internal {
+    function sim_reportExternalConsolidation(uint256 bufferedPrincipal, uint256 reportedAmount) internal {
         _simConsolidatedBalance += reportedAmount;
-        _simConsolidationRewardSurplus += consolidationRewardSurplus;
         _simTotalUserDeposited += bufferedPrincipal;
 
         if (reportedAmount > bufferedPrincipal) {
             _simCumulativeAutocompounded += reportedAmount - bufferedPrincipal;
         }
-        _simCumulativeAutocompounded += consolidationRewardSurplus;
     }
 
     /// @dev Convenience overload — delegates to the two-argument variant.
@@ -321,8 +310,8 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
             exitedArr[0] += exitedArr[i];
         }
 
-        // The consolidated amount and any queue-earned surplus have landed on the CL.
-        report.validatorsBalance = validatorsBalance + _simConsolidatedBalance + _simConsolidationRewardSurplus;
+        // The consolidated amount has landed on the CL.
+        report.validatorsBalance = validatorsBalance + _simConsolidatedBalance;
         report.validatorsSkimmedBalance = _simCumulativeSkimmed;
         report.validatorsExitedBalance = _simCumulativeExited;
         report.validatorsExitingBalance = validatorsExiting;
--- a/contracts/test/accounting/BeaconChainSimulator.sol
+++ b/contracts/test/accounting/BeaconChainSimulator.sol
@@ -51,11 +51,14 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
     /// @dev Cumulative ETH deposited on the EL deposit contract that has been activated on the CL.
     ///      Monotonically increasing — incremented in sim_activateValidators.
     uint256 internal _simTotalDepositedActivatedETH;
-    /// @dev Cumulative external-consolidation principal that has landed in validatorsBalance and been reported.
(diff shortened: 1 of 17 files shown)


contracts/test/accounting/BeaconChainSimulator.sol
```diff
@@ -228,6 +231,25 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
...
     }
 
+    /// @dev Models an external consolidation that will be included in the next oracle report.
+    ///      `bufferedPrincipal` is the amount minted into the consolidation buffer at request time.
+    ///      `reportedAmount` is the delta to add to the cumulative `totalExternalConsolidationsAmountReported` for this report.
+    ///      `validatorBalanceSurplus` is extra CL balance above `reportedAmount` that lands in validatorsBalance.
+    function sim_reportExternalConsolidation(
+        uint256 bufferedPrincipal,
+        uint256 reportedAmount,
+        uint256 validatorBalanceSurplus
```
juliaaschmidt: why do we need this extra surplus field here? Isn't the surplus the difference between `reportedAmount - bufferedPrincipal`. This sounds like a superfluous field

iamsahu: reportedAmount - bufferedPrincipal captures rewards included in the consolidation report. validatorBalanceSurplus captures additional rewards visible in validatorsBalance but not included in that consolidation amount—for example, rewards earned between consolidation and the oracle snapshot. We need both to model those two reporting layers independently.

juliaaschmidt: why would reportedAmount not include these extra rewards already? From my understanding the reportedAmount in the oracle report is the total current CL validators balance, which would also include the rewards earned between a consolidation and the oracle snapshot. 

if by `reportedAmount` you mean the increase in `totalExternalConsolidationAmountETH` reported then I get what you mean. If so could we please make the argument names less ambiguous?

iamsahu: That's right, will change it to make it more readable.

contracts/test/accounting/BeaconChainSimulator.sol
```diff
@@ -228,6 +231,26 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
         revert("sim_slash: no active validator found");
     }
 
+    /// @dev Models an external consolidation that will be included in the next oracle report.
+    ///      `bufferedPrincipal` is the principal minted into the consolidation buffer at request time (counted in
```
juliaaschmidt, line 231: By "principal minted" you mean amount in ETH?

contracts/test/accounting/BeaconChainSimulator.sol
```diff
@@ -228,6 +231,26 @@ abstract contract BeaconChainSimulator is AccountingHarnessBase {
...
+    ///      `_simTotalUserDeposited` for I2 ETH-conservation).
+    ///      `reportedAmount` is the delta to add to the cumulative `totalExternalConsolidationsAmountReported` for this report.
+    ///      `consolidationRewardSurplus` is extra CL balance above `reportedAmount` that lands in validatorsBalance (counted as rewards).
+    function sim_reportExternalConsolidation(
+        uint256 bufferedPrincipal,
+        uint256 reportedAmount,
```
juliaaschmidt: so if the `reportedAmount` isgreater than the `bufferedPrincipal` we got extra auto compounded rewards from the src 0x02??

However if `consolidationRewardSurplus`>0 it means that the source was an 0x01 validator that had some extra rewards? They would not be received by us. Similarly elFeeRecipient we would not assume that we would receive them

juliaaschmidt (approved):
LGTM

PR #652, opened 2026-07-07
feat(DepositDataBuffer Contract): DepositDataBuffer with batch-nonce ids and buffer-owned processed flag (BS-4393)

### LINKED ISSUE #546: feat(deposit/topups): Add the DepositDataBuffer contract and Attestation Contract to this repo
Currently the root attestation deposit flow relies on a data buffer contract and attestations from an attestation contract. Neither of which are in this repo. We should bring them in to be audit-complete. 

cc: @mischat @iamsahu

### LINKED ISSUE #548: Add a per-submission nonce to DepositObject so identical deposit batches don't permanently collide under replay protection
## Summary

The replay protection added in #537 turns a pre-existing edge case into a **permanent, hard block** for byte-identical deposit batches. The `depositDataBufferId` is cryptographically bound to the batch contents, and the `DepositObject` has no uniqueness field, so two submissions with identical `{pubkey, amount, operatorIdx}` entries collide on the same id — and after the first is marked processed, the second reverts `DepositDataBufferIdAlreadyProcessed` forever.

This is a **liveness/operational** issue — no fund loss, gated behind `onlyKeeper`.

## Where the id is bound to contents

- `contracts/src/interfaces/IDepositDataBuffer.sol:54-59` — `DepositObject` has only `Deposit[] deposits` and `TopUp[] topUps`; no nonce/salt/counter.
- `contracts/src/AttestationVerifier.1.sol:242-243` — enforcement:
  ```solidity
  bytes32 computedId = keccak256(abi.encode(batch));
  if (computedId != depositDataBufferId) revert BufferIdMismatch(depositDataBufferId, computedId);
  ```
- `contracts/src/interfaces/IDepositDataBuffer.sol:92-95` — `submitDepositData` requires the supplied id to equal `keccak256(abi.encode(batch))`.

So `depositDataBufferId` is a pure function of the batch contents — the producer cannot pick a fresh id for an otherwise-identical batch.

## NatSpec contradiction

`contracts/src/interfaces/IDepositDataBuffer.sol:8-10` currently instructs implementers:

> `depositDataBufferId` MUST be unique per submission — implementations should salt with an ever-incrementing counter (e.g. `lastQueueIdx++`).

This is **impossible to satisfy** as written: any salted id that isn't exactly `keccak256(abi.encode(batch))` reverts with `BufferIdMismatch`, and there is no field in the hash preimage to carry the counter.

## Concrete impact

A legitimate, economically-identical second top-up — e.g. "top up validator X by 1 ETH" submitted on two different days — produces the same `depositDataBufferId`. The second submission can never be executed as its own batch. Current workarounds (bundle into one batch, or co-include another entry to perturb the hash) are undocumented.

## Proposed fix

Add a batch-level `uint256 nonce` to `DepositObject`, populated with a random per-submission value (e.g. a UUIDv4 cast to `uint256`):

```solidity
struct DepositObject {
    Deposit[] deposits;
    TopUp[] topUps;
    /// @dev Random per-submission nonce (e.g. a UUIDv4 cast to uint256) that makes the
    ///      batch hash — and therefore depositDataBufferId — unique even when the
    ///      deposit/top-up contents are byte-identical to a prior submission.
    uint256 nonce;
}
```

Why this works and is low-risk:
- It's automatically included in `keccak256(abi.encode(batch))`, so a distinct `nonce` → distinct id. **No change needed in `AttestationVerifier`** — the `BufferIdMismatch` binding and attestation-over-id still hold (attesters sign the specific batch + nonce; not producer-forgeable).
- `uint256` is consistent with the struct's other scalar fields (`amount`, `operatorIdx`) and a single storage slot. A random 128/256-bit value is collision-proof in practice; the type also leaves the door open to a monotonic counter later without a type change.

### Tasks
- [ ] Add `uint256 nonce` to `DepositObject` (`IDepositDataBuffer.sol`).
- [ ] Reword the uniqueness NatSpec (`IDepositDataBuffer.sol:8-10`) to "salt with a unique random nonce (e.g. a UUIDv4)" instead of an incrementing counter.
- [ ] Add a test: two batches with identical `deposits`/`topUps` but different `nonce` produce different ids and both execute (and the same nonce still reverts `DepositDataBufferIdAlreadyProcessed`).
- [ ] **Off-chain (separate repo):** update the `DepositDataBuffer` producer to populate `nonce` with fresh randomness per submission. The on-chain change is necessary but not sufficient.

Found during the differential security review of #537 (finding M-1).

### LINKED ISSUE #621: feat: add river update call to `DepositDataBuffer` to mark `DepositDataBufferObject` as processed

juliaaschmidt:
## Summary

Ports **both** buffer contracts from [AlluvialFinance/frontrun-mitigation](https://github.com/AlluvialFinance/frontrun-mitigation): the `DepositDataBuffer` (from #32, BS-4393) — reconciled onto this repo's `IDepositDataBuffer` interface (the `deposits[]/topUps[]` DepositObject shape) and wired into the attestation deposit flow — and the standalone `AttestationBuffer` event relay.

Two improvements from #32 are folded in:
1. **Batch-nonce ids** — the buffer's `lastQueuedIdx` at submit time is folded into the id (`keccak256(abi.encode(batch, nonce))`), so byte-identical batches submitted more than once get distinct, individually-addressable ids.
2. **Buffer-owned `processed` flag** — a processor-gated `markDepositDataProcessed` lives on the buffer, and the buffer is the **sole authority** for replay protection.

## Issues closed

Closes #621 — on a successful `depositToConsensusLayerWithAttestation`, River makes a cross-contract call to the buffer (`markDepositDataProcessed`) to flip the batch's `processed` flag, and the deposit flow rejects any already-processed id (step 0 of `fetchAndValidateDeposits`), so a batch can never be handled twice.

Closes #548 — byte-identical deposit batches no longer permanently collide under replay protection.

Closes #546 — brings both the `DepositDataBuffer` and the `AttestationBuffer` (the attestation contract) into this repo, so the root-attestation deposit flow is audit-complete.

**Note on the #548 approach:** the underlying liveness problem is fixed, but via a different mechanism than the issue's proposed random `nonce` field on `DepositObject`. Instead of adding a struct field, the buffer assigns a **monotonic per-submission nonce** (`lastQueuedIdx`) and folds it into `depositDataBufferId = keccak256(abi.encode(batch, nonce))`; the nonce is stored and returned by `getDepositData` so the verifier can re-check the binding. `DepositObject` is left unchanged. The contradictory "salt with an ever-incrementing counter" natspec (which was impossible to satisfy) is corrected. Off-chain integration: the buffer writer computes the expected id from `lastQueuedIdx()` — no producer-supplied randomness required.

## Relationship to #645

This branch **supersedes #645** (which also targets #621). It includes the same River→buffer `markDepositDataProcessed` call, but takes the **buffer-sole-authority** design (deletes the verifier-side `ProcessedDepositDataBufferIds`) rather than #645's additive approach (verifier-side guard remains primary, buffer flag mirrors). The two designs are mutually exclusive. This branch also adds the full `DepositDataBuffer.sol` implementation and the nonce/interface reconciliation that #645 scoped out.

## Changes

- **`contracts/src/DepositDataBuffer.sol`** (new) — standalone, non-upgradeable. Writer-gated `submitDepositData` with nonce-folded id + field validation (pubkey/signature lengths, non-zero **gwei-aligned** amounts); processor-gated `markDepositDataProcessed`; admin-rotatable writer; stores the batch + nonce.
- **`IDepositDataBuffer.sol`** — `getDepositData` now returns `(DepositObject, uint256 nonce)`; added `markDepositDataProcessed`, `isDepositDataProcessed`, `getProcessor`, `lastQueuedIdx`, `setWriter`, the `DepositDataProcessed`/`SetWriter` events, and nonce on `DepositDataSubmitted`; kept `getWriter`/`getAdmin`; corrected the batch-uniqueness natspec (#548).
- **`AttestationVerifier.1.sol`** — step-0 replay check reads `buffer.isDepositDataProcessed`; tamper check recomputes `keccak256(abi.encode(batch, nonce))`. Attestation/EIP-712 signing is unchanged. Removed the verifier-side mark/is functions.
- **`ConsensusLayerDepositManager.1.sol`** — marks the batch processed on the buffer (via `verifier.getDepositDataBuffer()`), before external interactions.
- **Removed** `state/attestationVerifier/ProcessedDepositDataBufferIds.sol`.
- **Tests** — updated the three `IDepositDataBuffer` mocks; ported #32's unit + fuzz + invariant suites (`contracts/test/DepositDataBuffer.t.sol`, `contracts/test/fuzz/`, `contracts/test/invariant/`).
- **`contracts/src/AttestationBuffer.sol`** + **`contracts/src/interfaces/IAttestationBuffer.sol`** (new) — minimal, non-upgradeable, fully self-contained event relay (an L2 contract; **no** L1 `AttestationVerifier` coupling):
  - Happy path: `submitAttestation` emits `AttestationSubmitted(idx, depositDataBufferId, depositRootHash, signature)` and bumps `lastAttestationIdx`. Open submission; off-chain daemons collect the events and the `AttestationVerifier` performs ecrecover/quorum on the signatures at deposit time.
  - **Unhappy path (faulty batch):** there is **no** separate veto/error function. A committee member flags a batch as not-ready by submitting an ordinary attestation with `depositRootHash == 0` (a signature over `Attest(depositDataBufferId, 0)`). The buffer treats it like any other submission — it does no verification and holds **no per-id state**. Off-chain the signer is recovered and checked for committee membership exactly as for a normal attestation, so the flag is authenticated **by the signature**, not by any on-chain access control. Such a signal is **inert on L1**: `_verifyAttestationQuorum` requires the signed root to equal the live deposit root (never zero), so a zero-root attestation can never contribute to an actual deposit quorum. This replaces an earlier open `raiseError`/`vetoBatch` design whose sticky on-chain flag let **anyone** permanently make `submitAttestation` revert for a given id — a cheap DoS on the attestation path. Removing that flag eliminates the DoS by construction while keeping the buffer a pure, self-contained event relay.

## Design note

Replay authority now lives on the buffer, so replay history is tied to a specific buffer deployment — if the verifier's buffer address is swapped via `setDepositDataBuffer`, previously-processed ids on the old buffer don't carry over. This is the sole-authority tradeoff (documented in natspec). The deployed buffer's immutable processor must be the River proxy address.

## Testing

Full suite (excl. `HEAVY_FUZZING`) green (**1059 passing**). DepositDataBuffer suites: **34 unit + 7 fuzz + 3 invariant**; `DepositDataBuffer.sol` at **100%** coverage. AttestationBuffer suites: **8 unit + 2 invariant** (incl. an ecrecover-from-emitted-event round-trip and the `depositRootHash == 0` faulty-signal path); `AttestationBuffer.sol` at **100%** coverage. Verifier tests additionally prove a `depositRootHash == 0` attestation is attributable to a committee member yet cannot count toward a real-root quorum, and is inert when mixed into a valid quorum.

## Summary by CodeRabbit

* **New Features**
  * Added deposit buffering with queued batches, unique batch IDs, and a new attestation buffer for submissions.
  * Added support for tracking when deposit batches are processed.

* **Bug Fixes**
  * Strengthened replay protection so duplicate deposit batches can’t be accepted twice.
  * Batch IDs now include queue order, preventing identical deposits from colliding.
  * Added stricter validation for deposit amounts and input lengths.

* **Security**
  * Improved admin controls when changing the deposit buffer, including processor verification.


contracts/src/components/ConsensusLayerDepositManager.1.sol
```diff
@@ -163,8 +163,9 @@ abstract contract ConsensusLayerDepositManagerV1 is IConsensusLayerDepositManage
             depositDataBufferId, depositRootHash, signatures, depositContract, withdrawalCredentials, committedBalance
         );
 
-        // 5. Mark the batch ID processed BEFORE any external interactions.
-        verifier.markDepositDataBufferIdProcessed(depositDataBufferId);
+        // 5. Mark the batch ID processed on the buffer BEFORE any external interactions. The buffer
+        //    is the authoritative replay gate; `fetchAndValidateDeposits` reads this flag at step 0.
+        IDepositDataBuffer(verifier.getDepositDataBuffer()).markDepositDataProcessed(depositDataBufferId);
```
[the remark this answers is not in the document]
juliaaschmidt: Adding the init checks would mean a significant change to the protocol testing setup. if it was wrongly set, we have setter to set to another DepositDataBuffer. 

The processor on the DepositDataBuffer end is immutable by design (no proxy pattern), a new deposit data buffer would need to be deployed with the processor set to LC.

contracts/src/DepositDataBuffer.sol
```diff
@@ -0,0 +1,196 @@
...
+        if (depositCount == 0 && topUpCount == 0) revert EmptyDepositData();
+
+        // Amounts must be non-zero and gwei-aligned: the beacon-chain deposit contract encodes
+        // amounts in gwei, so a non-aligned amount would be silently truncated downstream. Rejecting
+        // here keeps the buffer consistent with the `InvalidDepositAmount` contract and the verifier.
+        for (uint256 i = 0; i < depositCount; i++) {
+            Deposit calldata d = batch.deposits[i];
+            if (d.pubkey.length != 48) revert InvalidPubkeyLength(i, d.pubkey.length);
+            if (d.signature.length != 96) revert InvalidSignatureLength(i, d.signature.length);
+            if (d.amount == 0 || d.amount % 1 gwei != 0) revert InvalidDepositAmount(i, d.amount);
```
iamsahu: Since we need the deposit amount to be at least 32 ETH & less than 2048 ETH we should add a check for that.

iamsahu (changes requested):
Apart from a minor change this looks good!

iamsahu (approved):

contracts/src/DepositAttestation.sol
```diff
@@ -0,0 +1,93 @@
...
+        _domainSeparator = domainSeparator;
+    }
+
+    /// @inheritdoc IDepositAttestation
+    function submitAttestation(
+        bytes32 depositDataBufferId,
+        bytes32 depositRootHash,
+        bytes calldata signature,
+        bytes calldata errorData
+    ) external {
```
zajck, line 49: Potential DoS vector, since anyone can submit attestation, even if they are not registered as attesters.

1. If keeper fails to validate that signatures belong to attesters, and submits them to river, they won't count towards the quorum and transaction will revert. If keeper is flooded with fake attestations, the real ones can never be properly processed.
https://github.com/liquid-collective/liquid-collective-protocol/blob/cc7318b3745a6fa9d246eb74e4f970dc045af4f7/contracts/src/AttestationVerifier.1.sol#L779-L797
2. Even if keeper properly filters out fake attestations, it's could be a demanding work.

If feasible, it would be good to gate this function to be called only by currently active attesters.

zajck: It was explained to me how the keeper works + since this contract is L2 contract, syncing the attesters across chains is more complex than just filtering the attestations.

zajck (approved):
👍

PR #665, opened 2026-07-23
test(pectra): assert SetOperatorExitedETH event index/value pairing

CWagamanEure:
## Summary

Closes coverage-review finding M-EV1: `SetOperatorExitedETH` was never `expectEmit`-checked, leaving the non-obvious wire mapping — event index `idx - 1`, value `_exitedETH[idx]` (array slot 0 is the total) — unguarded. A regression emitting `idx` instead of `idx - 1` would shift every operator's exited balance by one in off-chain indexers while all storage-based assertions stayed green.

- `testReportExitedETH` now asserts a `SetOperatorExitedETH(idx - 1, exitedETH[idx])` emission per operator before the `reportExitedETH` call, pinning the index/value pairing across fuzzed operator counts.
- Fixed the test-base event declaration to `uint256 indexed index`, matching `IOperatorRegistry.1.sol` — it was missing `indexed`, which would have made any `expectEmit` against it match the wrong topics.

## Testing

`forge test --match-test testReportExitedETH` — 7 passed, 0 failed (fuzz at 1500 runs).

### DIFF
--- a/contracts/test/OperatorsRegistry.1.t.sol
+++ b/contracts/test/OperatorsRegistry.1.t.sol
@@ -241,7 +241,7 @@ abstract contract OperatorsRegistryV1TestBase is Test {
 
     event SetRiver(address indexed river);
     event UpdatedExitedETH(uint256[] exitedETH);
-    event SetOperatorExitedETH(uint256 operatorIndex, uint256 exitedETH);
+    event SetOperatorExitedETH(uint256 indexed index, uint256 exitedETH);
 }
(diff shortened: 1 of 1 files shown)


juliaaschmidt (commented):
PR looks good. 👍 

The only thing is that for some reason codecov is failing now... is there any test we can add to increase the % codecov?

juliaaschmidt (approved):
Fixed it 👍

PR #676, opened 2026-08-10
feat(redemption-rewards): phase 2 — raise the redeem payout cap over [BS-4843]

iamsahu:
…stopped-earning demand

## Description

This pull request introduces a new "stopped earning" mechanism to the redeem manager, enabling the protocol to mark portions of redeem demand as having stopped accruing rewards at a fixed rate when their backing principal exits the consensus layer. This is achieved by tracking "rate marks" and "request anchors," which together allow for more accurate and fair payout calculations, especially in the presence of delayed withdrawals and protocol upgrades. The changes also ensure backward compatibility for requests made before this upgrade.

The most important changes are:

**Stopped Earning and Rate Mark Mechanism:**

* Added the `reportStoppedEarning` function to `RedeemManagerV1`, which marks redeem demand whose backing principal stopped earning, records a "rate mark" with the pre-report ETH/LsETH rate, and emits relevant events. [[1]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccR245-R299) [[2]](diffhunk://#diff-04225e1dec3c8a53677904ab85991a1183db03585363130dd35aab69f223cbfcR65-R79) [[3]](diffhunk://#diff-04225e1dec3c8a53677904ab85991a1183db03585363130dd35aab69f223cbfcR226-R254)
* Introduced `RateMarkStack` and `RedeemRequestAnchor` imports, storage, and accessors, along with new getter functions for retrieving rate mark and anchor details. [[1]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccR19-R20) [[2]](diffhunk://#diff-04225e1dec3c8a53677904ab85991a1183db03585363130dd35aab69f223cbfcR6-R7) [[3]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccR125-R139) [[4]](diffhunk://#diff-04225e1dec3c8a53677904ab85991a1183db03585363130dd35aab69f223cbfcR226-R254)

**Payout Calculation Logic:**

* Implemented internal helpers (`_rateMarkCursor`, `_settledHeight`, `_findRateMarkAtOrBefore`, `_sliceCap`) to calculate the ETH payout cap for redeem requests, taking into account any rate marks that may raise the cap above the original request-time value.
* Updated the redeem request claim logic to use the new anchor and rate mark data for post-upgrade requests, while maintaining legacy behavior for pre-upgrade requests. [[1]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccL427-R642) [[2]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccL447-R668) [[3]](diffhunk://#diff-27591a1fe04868ad6e4de630116d39ece6edb257b867d4e0ae99c1ad48074bccR561-R566)

**Oracle Reporting Integration:**

* Modified `LibOracleReporting` to value the stopped-earning principal at the pre-report rate, pass this data to the redeem manager, and ensure the stopped-earning report is always called before withdrawal settlement. [[1]](diffhunk://#diff-6851f7e96a4b96cd528f37ba4b22c3901ba83d0188d4b60e00c8560cbb79be7cR67) [[2]](diffhunk://#diff-6851f7e96a4b96cd528f37ba4b22c3901ba83d0188d4b60e00c8560cbb79be7cR209-R221) [[3]](diffhunk://#diff-6851f7e96a4b96cd528f37ba4b22c3901ba83d0188d4b60e00c8560cbb79be7cR378-R396)

These changes collectively add a robust mechanism for marking and handling stopped-earning principal, improving fairness and transparency for redeemers in the protocol.

## Notice

## Pull Request Type

## Testing

## Summary by CodeRabbit

- **New Features**
  - Added stopped-earning reporting to support accurate redemption settlement.
  - Redemption requests now use request-time valuation anchors and rate marks, including compatibility for legacy requests.
  - Added read-only access to rate-mark and redemption-anchor details.
  - Added notifications for stopped-earning reports and amounts exceeding markable demand.
  - Oracle reporting now includes stopped-earning LsETH values.

- **Documentation**
  - Clarified consolidation validation, fee refund handling, fund limits, and self-consolidation outcomes.


contracts/src/interfaces/IAttestationVerifierPectraMigration.1.sol
```diff
@@ -59,6 +59,13 @@ interface IAttestationVerifierPectraMigrationV1 {
     ///         post-Pectra paths (e.g. top-ups, normal consolidations). The lookup mutations
     ///         atomically revert with the rest of the transaction if the downstream
     ///         consolidation call fails.
+    /// @dev    That atomicity covers only the EL predeploy call, which merely QUEUES the request.
+    ///         Beacon-chain eligibility is evaluated later and an ineligible request is discarded
+    ///         with no on-chain signal, so the promotion can outlive a consolidation that never
+    ///         executed. The key is then recorded as post-Pectra while still holding 0x01
+    ///         credentials, which makes it top-up eligible via `fetchAndValidateDeposits`.
+    ///         Reconciliation is manual: the keeper calls `removeExitedValidatorPubkeys` and then
+    ///         `migratePrePectraValidatorPubkeys` to move it back.
```
juliaaschmidt: I don't follow the reasoning here. How would this would work `migratePrePectraValidatorPubkeys` should not register a previously (`removeExitedValidatorPubkeys`) exited key.

contracts/src/interfaces/IRedeemManager.1.sol
```diff
@@ -60,14 +64,31 @@ interface IRedeemManagerV1 {
...
+    /// @dev The re-pricing runs both ways. A `markedEth / amount` ratio below a covered request's
+    ///      request-time rate pushes that request's cap down. See `RedeemManagerV1._sliceCap`.
+    /// @param height The start position of the mark on the cumulative LsETH axis
+    /// @param amount The amount of LsETH marked
+    /// @param markedEth The ETH value of `amount` at the pool rate of this report
+    /// @param id The id of the new rate mark
+    event ReportedStoppedEarning(uint256 height, uint256 amount, uint256 markedEth, uint32 id);
+
+    /// @notice Emitted when reported stopped-earning principal exceeded the markable redeem demand
+    /// @dev Not an error. Most exits do not back a redemption, so the reported principal is usually far
```
juliaaschmidt, line 77: why does it say that an event is not an error here? 

Why does it say that the extra rewards do not belong to the redeemer here?

contracts/src/interfaces/IRedeemManager.1.sol
```diff
@@ -112,24 +133,36 @@ interface IRedeemManagerV1 {
...
     /// @notice Retrieve River address
     /// @return The address of River
     function getRiver() external view returns (address);
 
     /// @notice Retrieve the global count of redeem requests
+    /// @return The count of redeem requests
     function getRedeemRequestCount() external view returns (uint256);
 
     /// @notice Retrieve the details of a specific redeem request
+    /// @dev The `maxRedeemableEth` field of the returned struct is legacy-only: it bounds the payout for
```
juliaaschmidt: why don't we remove it then? to not change the queue?

contracts/src/state/redeemManager/RedeemRequestAnchor.sol
```diff
@@ -0,0 +1,46 @@
...
+///      claim settled below the request rate the implied ratio drifts upward without bound — see
+///      `testPartialClaimBelowRequestRateDriftsImpliedCapRate`, where 100 LsETH quoted at 1.0 leaves
+///      a residual implied rate of 50.5 ETH per LsETH.
+///
+///      The pair is stored rather than a precomputed rate so the cap arithmetic stays integer-exact
+///      and in the same pro-rata form the claim path already uses.
+///
+///      A zero `lsETHAtRequest` means the request predates this upgrade. Such requests are never
+///      marked and are always paid under the original rules, which is also how the launch cutover is
+///      enforced (the PRD excludes retroactive application).
```
juliaaschmidt: I don't think we'll have to mention the PRD here

contracts/src/AttestationVerifier.1.sol
```diff
@@ -659,6 +659,13 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1, IAttest
     ///        - `totalAmount` gwei alignment, upper bound, or correlation with the number
     ///          of (source, target) pairs.
     ///        - Whether the source validators actually exist on the consensus layer
+    ///        - Anything about `targetPubkeys` beyond byte length. In particular: a zero
+    ///          target is accepted, unlike a zero source which reverts; a target that is
+    ///          not a known funded River validator is accepted; and
```
juliaaschmidt: "a target that is not a known funded River validator is accepted;" since when?

contracts/src/AttestationVerifier.1.sol
```diff
@@ -659,6 +659,13 @@ contract AttestationVerifierV1 is Initializable, IAttestationVerifierV1, IAttest
...
+    ///          `sourcePubkeys[i] == targetPubkeys[i]` is accepted even though a self
```
juliaaschmidt: why is self consolidation allowed? This function is only used by `mintLsETHForConsolidation`

contracts/src/RedeemManager.1.sol
```diff
@@ -220,16 +236,71 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+        }
+
+        RedeemQueueV2.RedeemRequest[] storage redeemRequests = RedeemQueueV2.get();
+        uint256 requestCount = redeemRequests.length;
+        if (requestCount == 0) {
+            return;
+        }
+
+        // Claiming raises a request's height and lowers its amount by the same step, so its end position
+        // never moves. The last request's end position is therefore the total LsETH ever requested.
```
juliaaschmidt, line 265: if that's the case then why is this not asserted?

contracts/src/RedeemManager.1.sol
```diff
@@ -220,16 +236,71 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+        RedeemQueueV2.RedeemRequest storage lastRequest = redeemRequests[requestCount - 1];
+        uint256 totalRequestedHeight = lastRequest.height + lastRequest.amount;
+
+        // Marks may only cover demand that is still unsettled. A withdrawal event has already priced the
+        // slice below `settledHeight`, and that event's ETH caps what it pays out regardless. Marking it
+        // anyway would credit the redeemer with pool appreciation that accrued after their principal
+        // stopped earning.
+        uint256 markStart = _rateMarkCursor();
+        uint256 settledHeight = _settledHeight();
+        if (settledHeight > markStart) {
```
juliaaschmidt:
```suggestion
        if (settledHeight >= markStart) {
```
no?

contracts/src/RedeemManager.1.sol
```diff
@@ -220,16 +236,71 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+        }
+
+        uint256 reportedLsETH = _stoppedEarningLsETH;
+        uint256 lsETHToMark = reportedLsETH;
+        uint256 markable = totalRequestedHeight > markStart ? totalRequestedHeight - markStart : 0;
+        if (lsETHToMark > markable) {
+            lsETHToMark = markable;
+            emit StoppedEarningExceededMarkableDemand(reportedLsETH, lsETHToMark);
+        }
+        if (lsETHToMark == 0) {
```
juliaaschmidt: I feel that this check would make sense earlier. Either inside the previous if body (right after line 277) or right on line 273. It feels odd here.

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +311,183 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+
+    /// @notice Internal utility to find the last rate mark starting at or before a position
+    /// @dev The rate mark stack is sorted strictly ascending by height and non-overlapping, but unlike
+    ///      the withdrawal stack it has gaps, so `_performDichotomicResolution`'s contiguity assumption
+    ///      does not hold here. This is a plain predecessor search. The caller must still check whether
+    ///      the mark it returns reaches the position, or whether the position sits in a gap.
+    /// @param _height The position to search for
+    /// @return found True if any mark starts at or before `_height`
+    /// @return index The index of that mark
+    function _findRateMarkAtOrBefore(uint256 _height) internal view returns (bool found, uint256 index) {
```
juliaaschmidt, line 352: if the stack was contiguous we could have reused _performDichotomicResolution

juliaaschmidt (commented):
Some initial comments on the natspec changes

contracts/src/state/redeemManager/RedeemQueue.2.sol
```diff
@@ -11,7 +11,13 @@ library RedeemQueueV2 {
     struct RedeemRequest {
         /// @custom:attribute The amount of the redeem request in LsETH
         uint256 amount;
-        /// @custom:attribute The maximum amount of ETH redeemable by this request
+        /// @custom:attribute The maximum amount of ETH redeemable by this request. LEGACY: authoritative
```
juliaaschmidt: why do we keep it then?

contracts/src/RedeemManager.1.sol
```diff
@@ -220,16 +236,71 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+        // slice below `settledHeight`, and that event's ETH caps what it pays out regardless. Marking it
+        // anyway would credit the redeemer with pool appreciation that accrued after their principal
+        // stopped earning.
+        uint256 markStart = _rateMarkCursor();
+        uint256 settledHeight = _settledHeight();
+        if (settledHeight > markStart) {
+            markStart = settledHeight;
+        }
+
+        uint256 reportedLsETH = _stoppedEarningLsETH;
```
juliaaschmidt, line 279: this is a pure alias variable, unnecessary no?

contracts/src/interfaces/IRedeemManager.1.sol
iamsahu: Its AI fluff, we have some errors which has the exceed in them, for eg: `ExitsRequestedExceedExitDemand`.  Since this one has the word exceed and it is not an error it might have thought that it would be worth mentioning it. We can remove it.

It says that extra rewards or surplus doesn't belong to a redeemeer because we might over exit. For eg: to satisfy a demand of 1ETH we might exit a complete validator and which will lead to excess value of 31 ETH landing on the protocol.

contracts/src/state/redeemManager/RedeemRequestAnchor.sol
iamsahu: Agreed

contracts/src/AttestationVerifier.1.sol
iamsahu: This could happen while `mintLsETHForConsolidation`

contracts/src/AttestationVerifier.1.sol
iamsahu: Good catch, it should not be the case.

contracts/src/RedeemManager.1.sol
iamsahu: We can still come up with a refactor that will use a single function for doing the binary search for both the cases.

contracts/src/interfaces/IAttestationVerifierPectraMigration.1.sol
iamsahu: That's right. The reasoning is incorrect.

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +311,183 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+                }
+                cap += (unmarkedAmount * _anchor.ethAtRequest) / _anchor.lsETHAtRequest;
+                sliceCursor += unmarkedAmount;
+                remainingAmount -= unmarkedAmount;
+                // `markIndex` is not advanced. The candidate mark was not consumed and still applies to
+                // the new `sliceCursor`.
+                continue;
+            }
+
+            // case 2: the candidate mark ends at or below `sliceCursor`, so it covers no part of the slice
```
juliaaschmidt, line 458: this case has zero test coverage

contracts/src/RedeemManager.1.sol
```diff
@@ -414,18 +667,29 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
             vars.ethAmount =
                 (vars.matchingAmount * _params.withdrawalEvent.withdrawnEth) / _params.withdrawalEvent.amount;
 
-            // as each request has a maximum withdrawable amount, we verify that the eth amount is not exceeding this amount, pro rata
-            // the amount that is matched
-            uint256 maxRedeemableEthAmount =
-                (vars.matchingAmount * _params.redeemRequest.maxRedeemableEth) / _params.redeemRequest.amount;
+            // Each request carries a maximum withdrawable amount. Cap the eth pro rata to the amount
+            // matched here.
+            uint256 maxRedeemableEthAmount;
```
juliaaschmidt, line 687: The anchored branch computes the cap per slice, so unused request-time headroom no longer carries across withdrawal events the way the decrementing `maxRedeemableEth` budget did. 

Concretely, on a 30 LsETH request opened at 1.0, settled 20 LsETH at 0.5 then 10 LsETH at 1.5: legacy pays 10 + 15 = 25 ETH (old method), anchored pays 10 + 10 = 20 ETH with 5 ETH swept to BufferedExceedingEth (new logic in this PR). No rate mark is involved, so this applies to every post-upgrade request, not just marked ones. Test attached (passes on this branch as-is).

Was dropping the carry-forward a deliberate decision?

`testPartialClaimBelowRequestRateDriftsImpliedCapRate` reads like you consider this a bug, which would mean we want to revisit this? If it wasn't deliberate, the aggregate-over-the-request framing in the `_sliceCap` natspec is the thing to revisit, since the aggregation only holds within one withdrawal event.

```
//SPDX-License-Identifier: BUSL-1.1
pragma solidity 0.8.34;

import "./RedeemManager.1.t.sol";

/// @notice Candidate tests for PR #676, written in the style of the mark tests in
///         RedeemManager.1.t.sol. Paste the three members below into RedeemManagerV1Tests.
contract PR676CarryForward is RedeemManagerV1Tests {
    /// @dev Opens a 30 LsETH request at a rate of 1.0, then settles it in two withdrawal events and
    ///      claims after each: 20 LsETH priced at 0.5, then the remaining 10 LsETH priced at 1.5.
    ///      With `_clearAnchor` the request is made to look pre-upgrade, taking the legacy budget path.
    function _lowThenHighFill(uint256 _salt, bool _clearAnchor) internal returns (uint256 received) {
        address user = _generateAllowlistedUser(_salt);
        river.sudoSetRate(1e18);
        uint32 id = _openRequest(user, 30e18);
        assertEq(redeemManager.getRedeemRequestDetails(id).maxRedeemableEth, applyRate(30e18, 1e18));

        if (_clearAnchor) {
            bytes32 anchorSlot =
                keccak256(abi.encode(uint256(id), bytes32(uint256(keccak256("river.state.redeemRequestAnchor")) - 1)));
            vm.store(address(redeemManager), anchorSlot, bytes32(0));
            vm.store(address(redeemManager), bytes32(uint256(anchorSlot) + 1), bytes32(0));
            assertEq(redeemManager.getRedeemRequestAnchor(id).lsETHAtRequest, 0);
        }

        // no rate mark is ever pushed: the whole span sits in a gap and is capped at the request rate
        received = _settleAndClaim(id, 20e18, 0.5e18) + _settleAndClaim(id, 10e18, 1.5e18);
        assertEq(redeemManager.getRateMarkCount(), 0);
        assertEq(redeemManager.getRedeemRequestDetails(id).amount, 0);
    }

    /// The anchored cap is computed per slice from the request-time rate, so headroom left unspent by a
    /// fill that settled BELOW that rate is dropped rather than carried to the next fill. The cheap fill
    /// uses 10 of its 20 ETH slice cap; the expensive fill is worth 15 ETH but is capped at its own
    /// 10 ETH slice value, and the 5 ETH difference is confiscated.
    /// @dev Pairs with testLegacyCapCarriesUnusedHeadroomAcrossEvents, which is the behaviour this
    ///      replaces. No rate mark is involved, so this is the payout delta the anchor introduces on its
    ///      own, for every post-upgrade request.
    function testAnchoredCapDoesNotCarryUnusedHeadroomAcrossEvents() external {
        assertEq(_lowThenHighFill(0, false), applyRate(20e18, 0.5e18) + applyRate(10e18, 1e18));
        assertEq(redeemManager.getBufferedExceedingEth(), applyRate(10e18, 1.5e18) - applyRate(10e18, 1e18));
    }

    /// Legacy path, same two fills. `maxRedeemableEth` is one decrementing ETH budget for the whole
    /// request, so the 10 ETH the cheap fill did not spend is still on it when the expensive fill lands:
    /// that fill is paid in full at 15 ETH and nothing is confiscated.
    function testLegacyCapCarriesUnusedHeadroomAcrossEvents() external {
        assertEq(_lowThenHighFill(0, true), applyRate(20e18, 0.5e18) + applyRate(10e18, 1.5e18));
        assertEq(redeemManager.getBufferedExceedingEth(), 0);
    }
}
```

contracts/src/RedeemManager.1.sol
juliaaschmidt, line 1: Something worth flagging: 

A rate mark is a photograph of the price, taken at the moment your stake stopped earning on the beacon chain. Whatever price is in that photo becomes the ceiling on your payout. That's the feature's whole point, and normally it helps you: you asked to redeem at 1.0, the pool climbed to 1.05 while you waited in the exit queue, the photo caught 1.05, so you get 1.05 instead of the 1.0 you'd get today.

The catch is that a ceiling is a ceiling no matter what. The unfortunate scenario for a user:

 ```
Day 0   you request 30 LsETH, price 1.0        →  worth 30 ETH
Day 5   a validator gets slashed, price 0.9    →  pool is down
Day 6   your stake exits, photo taken @ 0.9    →  your ceiling is now 27 ETH
Day 20  coverage fund pays out, price back 1.0 →  pool is whole again
Day 25  your ETH is withdrawable, settles @ 1.0→  30 ETH is sitting there for you
 ```

 There are 30 ETH available for you on day 25. You get 27. The other 3 gets swept into BufferedExceedingEth and handed back to the people who didn't redeem. Today, with no photograph in the picture, you'd get all 30.

 That's the part I flagged: the coverage fund exists specifically to make people whole after a slashing. The recovery money travels to redeemers along exactly the route the new ceiling blocks. So the fund pays up, and the redeemer it was meant to protect doesn't see it.

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +311,183 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+    ///      the request has been pending across. A claimant pays for their own request's span and cannot
+    ///      be charged for anyone else's. There is no way to split the walk: `_depth` bounds the
+    ///      recursion across withdrawal events, not this loop, and `matchingAmount` is fixed by
+    ///      on-chain state. A request settled by a single large withdrawal event walks its whole span
+    ///      in one call. The only mitigation is claiming regularly.
+    /// @param _anchor The immutable request-time valuation of the request
+    /// @param _sliceStart The start position of the slice on the cumulative LsETH axis
+    /// @param _sliceAmount The amount of LsETH in the slice
+    /// @return cap The maximum ETH payable for this slice
+    function _sliceCap(RedeemRequestAnchor.Anchor memory _anchor, uint256 _sliceStart, uint256 _sliceAmount)
```
juliaaschmidt, line 401: Can we add a getter so a user can find out how much they will be paid out? 

Because `getRedeemRequestDetails` and `RequestedRedeem.maxRedeemableEth` are the fields the claim UI/relayer read today to anticipate the payout amount. This new method requires an off-chain `getRateMarkDetails` array. Is there a view we could add such as `previewClaimEth(redeemRequestId, withdrawalEventId)` or an external `getSliceCap` in this PR?

iamsahu: That's a good idea, I will add it.

contracts/src/RedeemManager.1.sol
iamsahu: Great point, will look into this.

contracts/src/RedeemManager.1.sol
iamsahu: This was to be revisited. Thanks for flagging it.

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +311,183 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+        uint256 remainingAmount = _sliceAmount;
+
+        // Marks are ascending and disjoint, so the only candidate that can cover `sliceCursor` is the last
+        // mark starting at or before it. Seek that one rather than scanning from the head of the stack.
+        (bool markFound, uint256 markIndex) = _findRateMarkAtOrBefore(sliceCursor);
+        if (!markFound) {
+            // The slice starts below every mark. Enter at the head of the stack and let the uncovered
+            // branch below value everything up to the first mark's start.
+            markIndex = 0;
+        }
```
zajck, line 425: It seems this if block is unnecessary, since when `markFound == false`, `markIndex` is 0 anyway.

juliaaschmidt:
@coderabbitai review

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +317,177 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+
+        while (remainingAmount > 0) {
+            if (markIndex >= markCount) {
+                // The walk has passed the last mark, so nothing can cover the rest of the slice. Value all
+                // of it at the request-time rate.
+                cap += (remainingAmount * _anchor.ethAtRequest) / _anchor.lsETHAtRequest;
```
zajck: ? What if the request-time rate is higher than withdrawal even rate?

contracts/src/RedeemManager.1.sol
```diff
@@ -240,9 +317,177 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+            // case 1: `sliceCursor` lies in the uncovered range below the candidate mark
+            if (sliceCursor < markStart) {
+                // No mark spans [sliceCursor, markStart), so no locked rate applies there and the interval
+                // keeps the request-time rate. A mark fixes the rate for a range, not the source of the
+                // ETH, since buffer ETH is fungible. Advance to min(markStart, sliceEnd).
+                uint256 unmarkedAmount = markStart - sliceCursor;
+                if (unmarkedAmount > remainingAmount) {
+                    unmarkedAmount = remainingAmount;
+                }
+                cap += (unmarkedAmount * _anchor.ethAtRequest) / _anchor.lsETHAtRequest;
```
zajck: What if the actual rate is below the rate at request?

contracts/src/RedeemManager.1.sol
```diff
@@ -435,23 +709,34 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+                _params.redeemRequest.amount -= vars.matchingAmount;
+            }
+            // Saturating subtraction. For a pre-upgrade request it is exact, because the cap above comes
+            // from this very field and can never exceed it. For a marked request the payout may exceed the
+            // request-time budget, and an unguarded decrement would revert the whole claimRedeemRequests
+            // call with Panic(0x11). Post-upgrade this field bounds nothing. The cap comes from the anchor
+            // and the rate marks.
+            _params.redeemRequest.maxRedeemableEth = _params.redeemRequest.maxRedeemableEth > vars.ethAmount
+                ? _params.redeemRequest.maxRedeemableEth - vars.ethAmount
+                : 0;
```
zajck, line 734: Why don't we just skip updating `_params.redeemRequest.maxRedeemableEth` for post-upgrade requests since it has no effect later?

contracts/src/RedeemManager.1.sol
```diff
@@ -362,7 +606,14 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
             })
         );
 
-        _setRedeemDemand(RedeemDemand.get() + _lsETHAmount);
+        // The request-time valuation, never mutated after this write. `maxRedeemableEth` cannot serve
+        // this purpose, because the claim path decrements it by the ETH actually paid, leaving an implied
+        // per-LsETH rate below the request rate once the request is partially claimed.
+        RedeemRequestAnchor.get()[redeemRequestId] =
+            RedeemRequestAnchor.Anchor({lsETHAtRequest: _lsETHAmount, ethAtRequest: maxRedeemableEth});
```
zajck, line 613: Since we store `maxRedeemableEth`, does it make sense to also store it in `redeemRequests`?

As far as I read it, the value `maxRedeemableEth` from `redeemRequests` has no affect anymore and no relevant meaning. Perhaps we can skip writing it to the storage (so it'll just be 0 for all new requests, but that should not be an issue, since already in the current implementation, it can later drop to 0, even when there are still live claims).

If we don't store it here, we could skip updating it's value in `_claimRedeemRequest`

contracts/src/RedeemManager.1.sol
iamsahu: As per the discussion the current approach is acceptable.

contracts/src/RedeemManager.1.sol
iamsahu: This has been fixed.

contracts/src/RedeemManager.1.sol
iamsahu: Good catch!

contracts/src/RedeemManager.1.sol
iamsahu: Fixed this.

contracts/src/RedeemManager.1.sol
iamsahu: Yup, removed it.

contracts/src/RedeemManager.1.sol
iamsahu: I have fixed the updation of the value in `_claimRedeemRequest`.

contracts/src/RedeemManager.1.sol
iamsahu: Fixed

contracts/src/RedeemManager.1.sol
```diff
@@ -435,23 +704,31 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
+                // cannot underflow. The end position it preserves is bounded by the total LsETH ever
+                // queued, so the increment cannot overflow.
+                _params.redeemRequest.height += vars.matchingAmount;
+                _params.redeemRequest.amount -= vars.matchingAmount;
+            }
+            if (_params.anchor.lsETHAtRequest == 0) {
+                _params.redeemRequest.maxRedeemableEth = _params.redeemRequest.maxRedeemableEth > vars.ethAmount
+                    ? _params.redeemRequest.maxRedeemableEth - vars.ethAmount
+                    : 0;
+            }
```
zajck, line 726:
```suggestion
            if (_params.anchor.lsETHAtRequest == 0) {
                unchecked {
                    _params.redeemRequest.maxRedeemableEth -= vars.ethAmount;
                }
            }
```
For legacy requests it's guaranteed that `_params.redeemRequest.maxRedeemableEth >= vars.ethAmount`

contracts/src/RedeemManager.1.sol
zajck: Would it make sense to change line 599 to 
```
maxRedeemableEth: 0
```
so if someone queries `getRedeemRequestDetails` and gets `maxRedeemableEth: 0`, it's an indicator they need to obtain the information in another way?

contracts/test/RedeemManager.1.t.sol
```diff
@@ -1649,11 +1668,534 @@ contract RedeemManagerV1Tests is RedeeManagerV1TestBase {
...
+    /// that fill is paid in full at 15 ETH and nothing is confiscated.
+    function testLegacyCapCarriesUnusedHeadroomAcrossEvents() external {
+        assertEq(_lowThenHighFill(0, true), applyRate(20e18, 0.5e18) + applyRate(10e18, 1.5e18));
+        assertEq(redeemManager.getBufferedExceedingEth(), 0);
+    }
+
+    /// The carry is an addend on top of `_sliceCap`, never a budget the mark uplift is drawn from. A fill
+    /// paid above the request rate because its span was marked must therefore leave the request-rate value
+    /// of the UNMARKED remainder fully payable: 15 LsETH at the locked 2.0 plus 15 at the request 1.0.
+    /// @dev This is the case a decrementing-budget fix gets wrong — there the 30 ETH uplift payout drains
```
zajck, line 1756: The `@dev` part is a bit confusing to me, since I don't know what "decrementing-budget fix" is (was) doing

contracts/src/state/redeemManager/RedeemQueue.2.sol
```diff
@@ -11,7 +11,13 @@ library RedeemQueueV2 {
...
+        ///                   only for requests created before the stopped-earning upgrade, the ones with
+        ///                   a zero `RedeemRequestAnchor`. For anchored requests the cap comes from the
+        ///                   anchor and the rate marks instead, this field is decremented with a
+        ///                   saturating subtraction, and it can reach 0 while the request is still
+        ///                   claimable. See `RedeemRequestAnchor` for why this field cannot carry the
```
zajck, line 19: Since the subtraction in `_claimRedeemRequest` was changed, the description is not accurate anymore.

contracts/src/RedeemManager.1.sol
iamsahu: I was working on the following to address it: https://github.com/liquid-collective/liquid-collective-protocol/pull/691

contracts/test/RedeemManager.1.t.sol
iamsahu: The decrementing-budget is an implementation using one decreasing payout budget for the entire request.

I will remove it so that it doesn't create confusion in the future.
Thanks for raising this!

contracts/src/RedeemManager.1.sol
iamsahu: Will incorporate.

contracts/src/RedeemManager.1.sol
iamsahu: I have gone ahead with #691 to preserve what APIs expect.

contracts/src/state/redeemManager/RedeemQueue.2.sol
iamsahu: Fixed it

zajck (approved):
LGTM 👍

PR #677, opened 2026-08-10
feat(redemption-rewards): phase 3 — pin the launch cutover with a rat…

iamsahu:
…e mark floor

Closes a gap phase 2 left open. A zero RedeemRequestAnchor already makes a pre-upgrade request ignore rate marks when its payout is capped, which satisfies the PRD's exclusion of retroactive application on the paying side. But ignoring a mark is not the same as not consuming one: marks advance a single cursor across the queue, so the first reports after the upgrade would spend their stopped-earning credit covering pre-upgrade requests that cannot use it, and the first post-upgrade cohort would be short-changed by exactly that amount.

RateMarkFloor is the lowest position a mark may cover, set once by initializeRedeemManagerV1_3 to the total LsETH ever requested at upgrade time — the end of the pre-upgrade queue. reportStoppedEarning now takes max(markCursor, settledHeight, floor). testFloorStopsLegacyDemandFromConsumingMarks is the regression: a legacy request settles at its request rate as before, AND the post-upgrade request still receives the whole mark.

initializeRedeemManagerV1_3 is init(2), which is correct for a live deployment: mainnet has already run initializeRedeemManagerV1 and initializeRedeemManagerV1_2, so its version is 2. It must run in the same governance action as the implementation upgrade — otherwise the floor stays 0 and the first reports leak credit as described above.

Worth recording for whoever writes the next migration: RedeemQueueV1 and RedeemQueueV2 share the storage slot keccak256("river.state.redeemQueue") - 1, so _redeemQueueMigrationV1_2 re-interprets that array in place. It is safe only because init(1) gates it to exactly one run before any V2 request exists; invoking it over a populated V2 queue corrupts every element. The test helper hit this and now pokes the version slot instead, with the reason documented at the call site. V1_3 adds no queue migration at all — the anchor and the floor live at their own slots — so it cannot repeat the hazard.

RedeemManagerV1 11,108 -> 11,417 runtime bytes (13,159 free). RiverV1 unchanged. 5 new tests; full suite 1075 pass.

## Notice

## Pull Request Type

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added an upgrade path that establishes a boundary for rate marks.
  * Added visibility into the configured rate-mark boundary.

* **Bug Fixes**
  * Prevented older stopped-earning redemption requests from consuming rate marks created after the upgrade.
  * Improved redemption accounting when marking and settlement occur in the same oracle report.
  * Preserved locked mark values and accurate payouts across split, capped, and complex redemption scenarios.


contracts/src/RedeemManager.1.sol
```diff
@@ -68,10 +69,27 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
...
     function initializeRedeemManagerV1_2() external init(1) {
         _redeemQueueMigrationV1_2();
     }
 
+    /// @inheritdoc IRedeemManagerV1
+    function initializeRedeemManagerV1_3() external init(2) {
+        // Pin the launch cutover for stopped-earning accrual at the end of the existing queue, so the
+        // requests already pending at upgrade time neither accrue (they have no anchor) nor consume the
+        // marks that the first post-upgrade cohort is owed.
+        RedeemQueueV2.RedeemRequest[] storage redeemRequests = RedeemQueueV2.get();
```
juliaaschmidt, line 83: this whole block is repeated code here: `reportStoppedEarning` (lines 275-282)

We could have a shared helper function:
``` 
 /// @return The total LsETH ever requested
    function _totalRequestedHeight() internal view returns (uint256) {
```

contracts/src/RedeemManager.1.sol
```diff
@@ -269,6 +292,12 @@ contract RedeemManagerV1 is Initializable, ReentrancyGuard, IRedeemManagerV1, IP
         if (settledHeight > markStart) {
             markStart = settledHeight;
         }
+        // Never mark below the launch cutover. Pre-upgrade requests cannot use a mark, so letting the
+        // cursor cover them would burn credit owed to the first post-upgrade cohort.
+        uint256 floor = RateMarkFloor.get();
+        if (floor > markStart) {
```
juliaaschmidt, line 304: `markStart = floor` relocates the delta rather than dropping it — that's the bug.

The branch only fires while the pre-upgrade queue is still draining (`floor > max(cursor, settledHeight`)). In that window the reported principal has exited to serve legacy demand, but the line pushes the mark up to floor anyway, minting that accrual onto the first post-upgrade requests at drain-era rates and advancing `_rateMarkCursor()` past their span. Later, when those requests' own principal crosses `exit_epoch`, `markable` is already 0, so their real credit is dropped.

Net effect: the launch cohort has its accrual locked to whatever rate happened to prevail while the legacy queue drained, not the rate at which their backing actually stopped earning. The longer the drain, the worse the mismatch — so it lands squarely on the cohort the floor was added to protect.

Probe: a request opened at 1.0 whose principal stops earning at 1.5 is paid 10 ETH instead of 15, with the 5 ETH swept to `BufferedExceedingEth` (`StoppedEarningExceededMarkableDemand(10e18, 0)` on the second report, `getRateMarkCount() == 1`).

The comment at :295-296 only reasons about legacy demand consuming marks which this correctly prevents, but not about legacy exits minting them. If the intent is that legacy-driven exits generate no entitlement at all, `return`ing instead of clamping is the one-line fix. We should think through these edge cases and clarify our intended behavior

juliaaschmidt:
@coderabbitai review

contracts/src/RedeemManager.1.sol
juliaaschmidt: I tried to illustrate this [here](https://github.com/liquid-collective/liquid-collective-protocol/pull/690/changes#top)

zajck: I addition to what @juliaaschmidt wrote, I think there's another issue here. 

Even if we correctly drop all rate marks that should have no effect, we will eventually come to the situation when the first meaningful rate mark will come in. But then it's possible that this relocation moves the whole rate mark higher, while only part of it should affect the rate.

Scenario: suppose that at the upgrade time, the redeem queue height is 10, so the floor will be 10.
Then at some point later the withdrawal stack height is 9. And in between the redeem queue increased to 15. At that point, stopped earning report comes in, reporting 4 eth that stopped working.

Normally, I'd expect markStart would be put to the withdrawal height 9, and user, whose redemption request starts at 10, would get 3 LSeth priced at rate mark and 2 at the max withdrawable rate set at the request time. 
But in the current code, the user would get 4 LSeth priced at mark rate and 1 at max withdrawable rate.

Maybe this is just a variation if what Julia reported, I suppose the difference is that the correct rate mark is used, but just at the wrong height.

iamsahu: Great catch, I will push a fix for this.

contracts/src/RedeemManager.1.sol
iamsahu: This has been removed now.

contracts/test/RedeemManager.1.t.sol
```diff
@@ -2198,6 +2217,1120 @@ contract RedeemManagerV1Tests is RedeeManagerV1TestBase {
...
+
+    /// _claimRedeemRequest's maxRedeemableEth decrement is saturating, not checked, because a
+    /// marked payout may legitimately exceed the request-time ETH budget: once the field floors at
+    /// 0, a checked subtraction on a LATER fill of the same request would revert the entire
+    /// claimRedeemRequests call with Panic(0x11). Every existing marked-request test claims in one
+    /// shot, so this post-saturation state -- and the fact that the cap keeps paying the mark rate
+    /// afterwards, recomputed from the anchor and the marks rather than from the exhausted budget
+    /// -- was never observed. The second claim below succeeding at all is the regression guard: a
+    /// reintroduced checked decrement would revert it with Panic(0x11).
+    function testMultiFillOfMarkedRequestSaturatesMaxRedeemableEthWithoutReverting() external {
```
zajck, line 2587: This test is not relevant since calculation in `_claimRedeemRequest` was changed.
So it could either be removed, or repurposed (i.e. testing that `redeemManager.getRedeemRequestDetails(id).maxRedeemableEth` stays constant)

iamsahu: I will remove it.

zajck (approved):
LGTM

PR #680, opened 2026-08-24
refactor(attestation-verifier): replace onlyRiverAdmin with an init-time admin

iamsahu:
## Motivation

`AttestationVerifierV1` gated its 8 privileged functions behind `onlyRiverAdmin`, which resolved governance through a cross-contract call:

```solidity
modifier onlyRiverAdmin() {
    if (msg.sender != IAdministrable(RiverAddress.get()).getAdmin()) revert LibErrors.Unauthorized(msg.sender);
    _;
}
```

Problem with that coupling:

- **Runtime dependency on River** Already overloaded River would have to include functions to be able to call AttestationVerifier
- **No way for the River Admin (Firewall) to call the AttestationVerifier**
## Change

The verifier now owns its admin. It inherits [`Administrable`](contracts/src/Administrable.sol) — the same pattern as `OperatorsRegistryV1`, `OracleV1`, and `AllowlistV1` — giving it `onlyAdmin`, `getAdmin()`/`getPendingAdmin()`, and the two-step `proposeAdmin`/`acceptAdmin` handover for free. The admin lives in the shared `AdministratorAddress` slot, whose `set()` already rejects the zero address via `LibSanitize._notZeroAddress`, so a misconfigured deploy cannot leave the verifier ungoverned.

River's address is still stored — it remains the EIP-712 `verifyingContract` anchor and the sole authorized caller of the `onlyRiver` entry points. Only River's *admin* loses rights here. Deployments can pass the same governance address as `_admin` to preserve today's operational setup.

No new storage library, modifier, or error was needed. Deployed size stays well under EIP-170 (~17.5 KB of 24.576 KB).

## ⚠️ Breaking ABI change

`initAttestationVerifierV1` takes `_admin` as its **first** parameter:

```solidity
function initAttestationVerifierV1(
    address _admin,          // <-- new
    address _river,
    address _depositDataBuffer,
    address[] calldata _rootAttesters,
    uint256 _quorum,
    bytes4 _genesisForkVersion,
    address[] calldata _consolidationCommitteeAttesters,
    uint256 _consolidationQuorum
) external;
```

Deployment tooling must be updated. The contract is **not yet deployed** (nothing in `deploy/`, `deployments/`, `hardhat_scripts/`, or `deployment.mainnet.json` references it), so there is no storage-migration or upgrade-compatibility concern and no migration path to worry about.

## Tests

New [`contracts/test/components/AttestationVerifierAdmin.t.sol`](contracts/test/components/AttestationVerifierAdmin.t.sol) covers:

- init stores the admin and emits `SetAdmin`; init reverts `InvalidZeroAddress` on a zero admin
- table-driven `Unauthorized(caller)` assertion across all 8 `onlyAdmin` functions, and the mirror case that the admin always clears the gate
- two-step transfer: `proposeAdmin` is `onlyAdmin`, `acceptAdmin` is `onlyPendingAdmin`, pending resets on acceptance, old admin loses access
- **regression guard 1** — River's admin is `Unauthorized` on every admin function (would fail if `onlyRiverAdmin` came back)
- **regression guard 2** — admin setters work with *no code* at the River address, proving the cross-contract lookup is gone

Existing call sites updated across `River.1.t.sol`, `AccountingHarnessBase.sol`, `ConsensusLayerDepositManagerAttestation.t.sol`, `ConsolidationAttestation.t.sol`, `SetDepositDomainFromForkVersion.t.sol`, and `AttesterSetInvariantTest.t.sol`, dropping the now-dead River-admin stubs and `vm.mockCall(river, getAdmin(), ...)` wiring.

`forge test --no-match-contract HEAVY_FUZZING`: **1019 passed**, 1 failed — `TLC_globalUnlockScheduleMigration.t.sol`, a pre-existing fork test that needs `MAINNET_FORK_RPC_URL`, unrelated to this change. `forge build` and `forge fmt --check` both clean.


zajck (dismissed):
LGTM 👍

zajck (approved):

juliaaschmidt (approved):
LGTM

@iamsahu However we should also deploy and add an AttestationProxyFirewall to the deployment scripts then?
 `_admin` must be a new `AttestationVerifierFirewall (Firewall(governor, executor, verifierProxy, [selectors…]))`, not the governance multisig directly. The repo's convention is one Firewall per Administrable contract (RiverFirewall, OracleFirewall,
 OperatorsRegistryFirewall, AllowlistFirewall, RedeemManagerProxyFirewall).

PR #688, opened 2026-09-03
chore: deposit tests with real BLS signatures [BS-5016]

zajck:
## Description

Tests were mocking BLS signature, since EIP-2537 precompiles weren't enabled in Foundry by default.
They are now, so this PR adds a signing utility that creates a valid BLS signature.

-  Added new util `BLSSigner.sol`, which signs the deposit data. It reuses `BLS12_381.sol` where possible. Note that it needs to call `BLS12_381.hashToG2` through a self-staticcall trampoline, since calling it directly multiple times in a single call results in a dirty scratchpad and wrong hash.
- `ConsensusLayerDepositManagerAttestation.t.sol` is updated to use actual signatures. New test were added, making sure wrong BLS signatures fail.
- `River.1.t.sol` and `AccountingHarnessBase.sol` still use mock `verifyBLSDeposit` call since the checks in those tests do not depend on valid BLS signatures. Only the comment before mocking was updated.

## Notice

## Pull Request Type

## Testing

### DIFF
--- a/contracts/test/components/ConsensusLayerDepositManagerAttestation.t.sol
+++ b/contracts/test/components/ConsensusLayerDepositManagerAttestation.t.sol
@@ -926,6 +926,192 @@ contract ConsensusLayerDepositManagerAttestationTest is Test {
     }
 
     // -----------------------------------------------------------------------
+    // Invalid BLS signature tests.
+    //
+    // Every case below is a batch the root attesters legitimately signed off on — the quorum is
+    // valid and the buffer id matches — so the only thing standing between it and a real
+    // consensus-layer deposit is verifyBLSDeposit. Each must revert the whole transaction.
+    //
+    // The batch is always tampered with *before* _prepareDeposit, because the buffer id commits
+    // to the batch contents: mutating a deposit afterwards would trip BufferIdMismatch first and
+    // the BLS path would never be reached.
+    // -----------------------------------------------------------------------
+
+    /// @dev A signature that is a valid BLS signature, just by the wrong key. Both the signature
+    ///      and its Y coordinate come from key B, so the point is well-formed and on-curve and the
+    ///      pairing check runs to a clean `false` rather than failing the precompile.
+    function testRevert_blsRejectsSignatureFromDifferentKey() public {
+        IDepositDataBuffer.Deposit[] memory deposits = new IDepositDataBuffer.Deposit[](1);
+        deposits[0] = _makeDeposit(0, 1001);
+
+        IDepositDataBuffer.Deposit memory other = _makeDeposit(0, 1002);
+        deposits[0].signature = other.signature;
+        deposits[0].depositY.signatureY = other.depositY.signatureY;
+
+        (bytes32 bufferId, bytes32 rootHash, bytes[] memory sigs) = _prepareDeposit(deposits);
+
+        vm.prank(keeper);
+        vm.expectRevert(BLS12_381.InvalidSignature.selector);
+        dm.depositToConsensusLayerWithAttestation(bufferId, rootHash, sigs);
+    }
+
+    /// @dev The amount is part of the signed deposit message, so a signature lifted from a 32 ETH
+    ///      deposit must not authorise a 64 ETH one. Guards against an operator inflating a
+    ///      deposit while reusing a previously valid signature.
+    function testRevert_blsRejectsSignatureOverDifferentAmount() public {
+        IDepositDataBuffer.Deposit[] memory deposits = new IDepositDataBuffer.Deposit[](1);
+        deposits[0] = _makeDepositWithAmount(0, 1003, 32 ether);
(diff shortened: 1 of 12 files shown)


contracts/test/utils/BLSSignerValidation/BLSFlagEdgeCases.t.sol
```diff
@@ -0,0 +1,348 @@
...
+        returns (bytes32)
+    {
+        return BLS12_381.depositMessageSigningRoot(pubkey, amount, wc, domain);
+    }
+}
+
+/**
+ * @title BLSFlagEdgeCasesTest
+ * @notice Solidity counterparts to the `BLS flag edge cases` block of noble-curves'
+ *         `test/bls12-381.test.ts`, plus its infinity-point and non-canonical-limb cases.
```
juliaaschmidt, line 53: Sounds like this was taken from a typescript project? l.53-l.68 if we adapt work from other projects, we should cite them on the top of each file.

contracts/test/utils/BLSSignerValidation/BLSVectors.sol
juliaaschmidt, line 1: if this is auto-generated file we should move all the generated files into a `generated/` path or some other rule to make it clear.

scripts/bls-gen-vectors.js
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env node
+//
+// Regenerates contracts/test/utils/BLSSignerValidation/BLSVectors.sol.
+//
+// Two sources, deliberately:
+//   1. The `sign` handler of ethereum/bls12-381-tests (CC0-1.0) — official, authoritative, and
```
juliaaschmidt, line 6: The licensing is confusing. What licensing is this file? We put  BUSL-1.1 on the other new files but this one mentions CC0-1.0 fixture data.

contracts/test/components/ConsensusLayerDepositManagerAttestation.t.sol
```diff
@@ -934,6 +925,192 @@ contract ConsensusLayerDepositManagerAttestationTest is Test {
...
+    //
+    // The batch is always tampered with *before* _prepareDeposit, because the buffer id commits
+    // to the batch contents: mutating a deposit afterwards would trip BufferIdMismatch first and
+    // the BLS path would never be reached.
+    // -----------------------------------------------------------------------
+
+    /// @dev A signature that is a valid BLS signature, just by the wrong key. Both the signature
+    ///      and its Y coordinate come from key B, so the point is well-formed and on-curve and the
+    ///      pairing check runs to a clean `false` rather than failing the precompile.
+    function testRevert_blsRejectsSignatureFromDifferentKey() public {
```
juliaaschmidt, line 943: Finally some great tests in this file 🥳

package.json
```diff
@@ -10,9 +10,12 @@
     "compile": "hardhat compile",
     "hh": "hardhat",
     "uml": "sol2uml ./contracts/src -o ./docs/components.svg",
-    "link_contracts": "cd node_modules && rm -rf openzeppelin-contracts && ln -s ../lib/openzeppelin-contracts && rm -rf openzeppelin-contracts-upgradeable && ln -s ../lib/openzeppelin-contracts-upgradeable"
+    "link_contracts": "cd node_modules && rm -rf openzeppelin-contracts && ln -s ../lib/openzeppelin-contracts && rm -rf openzeppelin-contracts-upgradeable && ln -s ../lib/openzeppelin-contracts-upgradeable",
+    "bls:vectors": "node scripts/bls-gen-vectors.js",
+    "bls:crosscheck": "node scripts/bls-crosscheck.js"
   },
   "dependencies": {
+    "@noble/curves": "1.4.2",
```
juliaaschmidt, line 18: technically a dev dependency only, also why the old version? https://www.npmjs.com/package/@noble/curves 2.4.0 is the latest

iamsahu (dismissed):
LGTM! 🚀 

It would be great to have all references featured prominently at the top of the files wherever needed.

juliaaschmidt (dismissed):

package.json
juliaaschmidt: just can hear renovatebot yelling at us already

scripts/bls-gen-vectors.js
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env node
 //
-// Regenerates contracts/test/utils/BLSSignerValidation/BLSVectors.sol.
+// This script is licensed like the rest of the repo (BUSL-1.1, see LICENSE). It downloads and
+// embeds fixture data from ethereum/bls12-381-tests, which that project licenses CC0-1.0 — that
```
juliaaschmidt, line 4: 🔥

juliaaschmidt (approved):
LGTM

iamsahu (approved):
LGTM!

